PR #4, opened 2021-03-20
Fixed setup.py and added a simple test file which can be run to check…

### LINKED ISSUE #3: Installation with pip
If I do:
```
git clone git@github.com:ExaWorks/psi-j-python.git
virtualenv -p python3 ./.virtualenv
source .virtualenv/bin/activate
pip install -r requirements.txt
pip install -e .
```

And then try to run one of the example files in `test`, I get:
```
(126) herbein1 ~/Repositories/exaworks/psi-j-python/tests [.virtualenv] (main ?)
❯ python ./test_doc_examples.py                                                                                                                                                              17:08:01 ()
Traceback (most recent call last):
  File "/Users/herbein1/Repositories/exaworks/psi-j-python/tests/./test_doc_examples.py", line 1, in <module>
    import psi.j
ModuleNotFoundError: No module named 'psi'
```

Looking at the virtualenv, it seems the package was installed as `jpsi-python`:
```
❯ find ./.virtualenv -name "*psi*"                                                                                                                                                           17:08:37 ()
./.virtualenv/lib/python3.9/site-packages/jpsi-python.egg-link
```

I don't totally know my way around the packaging of python, so I'll defer to others, but I think the issue here is in `setup.py` which has the name of the package as `jpsi-python`.   Not sure if we want to tweak the `setup.py` to have a different package name or tweak the imports.

hategan:
… basic functionality

Fixes #3 

To check:
```sh
pip3 install -e .
python3 tests/run_local_job.py
```

Should print something like

```
Job submitted
Job done: JobStatus[COMPLETED, time=1616205826.194412, exit_code=0]
```

### DIFF
--- a/tests/run_local_job.py
+++ b/tests/run_local_job.py
@@ -1,6 +1,6 @@
 # This is meant as a simple test file to check if psi/j was installed successfully
 
-from psi.j import SubmitException, Job, JobExecutor, JobSpec, JobState
+from psi.j import Job, JobExecutor, JobSpec
 
 
 if __name__ == '__main__':
--- a/tests/run_local_job.py
+++ b/tests/run_local_job.py
@@ -1,0 +1,12 @@
+# This is meant as a simple test file to check if psi/j was installed successfully
+
+from psi.j import SubmitException, Job, JobExecutor, JobSpec, JobState
+
+
+if __name__ == '__main__':
+    job = Job(JobSpec(executable='/bin/date'))
+    exec = JobExecutor.get_instance('local')
+    exec.submit(job)
+    print('Job submitted')
+    status = job.wait()
+    print('Job done: {}'.format(status))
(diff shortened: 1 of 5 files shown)


tests/run_local_job.py
```diff
@@ -0,0 +1,12 @@
+# This is meant as a simple test file to check if psi/j was installed successfully
+
+from psi.j import SubmitException, Job, JobExecutor, JobSpec, JobState
```
andre-merzky, line 3: Some imports are not used which probably will trip a linter.

andre-merzky (approved):
LGTM!

tests/run_local_job.py
hategan: Good point. The tests were not even checked.

I added the tests dir to "make checks" and removed the unused imports.

SteVwonder (approved):
LGTM!

PR #5, opened 2021-03-25
Use str instead of a specialized class for native ids (although allow…

hategan:
… Job to maintain the native id in another representation)

I'm still a bit fuzzy whether this is the right solution or whether we should add to/from str methods to NativeID.

### DIFF
--- a/src/psi/j/__init__.py
+++ b/src/psi/j/__init__.py
@@ -16,14 +16,13 @@ from .job_spec import JobSpec
 from .job_state import JobState
 from .job_status import JobStatus
 from .launchers.launcher import Launcher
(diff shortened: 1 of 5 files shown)


andre-merzky (approved):
> I'm still a bit fuzzy whether this is the right solution or whether we should add to/from str methods to NativeID.

I don't think a `NativeID` type adds value to the user of the API.

hategan:
> I don't think a `NativeID` type adds value to the user of the API.

Agreed. It would mostly make sense from a "proper design" design perspective.

SteVwonder (approved):
LGTM too!

PR #6, opened 2021-03-30
Create python-package.yml

hategan:
Adds github workflow for running checks and tests on multiple python versions

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -25,8 +25,6 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
-        python -m pip install flake8 pytest
         pip install -r requirements.txt
         pip install -r requirements-dev.txt
     - name: Lint with flake8
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -29,9 +29,6 @@ jobs:
         python -m pip install flake8 pytest
         pip install -r requirements.txt
         pip install -r requirements-dev.txt
-    - name: Build scripts
-      run: |
-        make launcher-scripts
     - name: Lint with flake8
       run: |
         make checks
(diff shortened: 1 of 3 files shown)


andre-merzky (approved):
LGTM

andre-merzky (commented):
Alas, it does not yet work - but I guess you'll iterate on this PR towards that :-)

hategan:
Allright, should be good to go

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install flake8 pytest
```
SteVwonder, line 29: Should `flake8` and `pytest` be included in `requirements-dev.txt`?

.github/workflows/python-package.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
```
SteVwonder, line 18: Curious as to why 3.6 was dropped.  Is there anything in particular in 3.6 that is causing issues?

.github/workflows/python-package.yml
hategan: Done.

.github/workflows/python-package.yml
hategan: There was an error. I think it complained about "from __future__ import annotations" in executors/local.py. So I removed 3.6 because deciding on the exact python support matrix was not the purpose of this PR.

.github/workflows/python-package.yml
SteVwonder: Great!

Nit: I think this line can be dropped from the github action then, since the `pip install -r requirements-dev.txt` now covers it.

SteVwonder (approved):
LGTM!  @hategan: feel free to merge now if you want.  We can always drop the extra pip line down the road.  No need for that to artificially block this PR.

PR #8, opened 2021-03-30
Add __repr__ for _ProcessEntry, so that it's informative when logged,…

hategan:
… and move the process property to the base class.


andre-merzky (approved):
LGTM

PR #9, opened 2021-03-30
When job output is redirected, a file is opened for that purpose. A m…

hategan:
…ethod existed to close those files, but was not properly called. This commit fixes that.


andre-merzky (approved):
Looks reasonable.

PR #10, opened 2021-03-30
The version lookup code for executors was broken. This commit fixes it.

hategan:


andre-merzky (approved):
I wasn't able to fully appreciate the changes, but don't see anything obviously wrong at least

hategan:
> I wasn't able to fully appreciate the changes, but don't see anything obviously wrong at least

I wasn't either, until I ran into the previous issue with having the same path twice in sys.path and the code was trying to compare the "new" version with the "existing" version.

You can test this by disabling the duplicate sys.path entry removal in __init__.py, then adding the src dir twice to PYTHONPATH and finally attempting to get an instance of the local executor.

PR #11, opened 2021-03-30
Fix launchers

### LINKED ISSUE #1: Implement pre/post-launch
Currently, launchers transform a spec into a list of arguments that are meant to be executed by executors, either directly, or placed into a submit script.

But this does not leave proper room for pre-launch and post-launch scripts, since they need to be "sourced". This implies that:
1. They must be "sourced" in a script
2. That script must be a script that also contains the launch command.

This is not currently straightforward with the local executor and the single launcher since no script is generated by either.

Things should change such that:
1. All launchers use a script.
2. Pre and post launch scripts are sourced by the launcher scripts.

hategan:
This should fix #1. The basic idea if this PR is:
- use (bash) scripts for all launchers
    - there is common code in the scripts, which is stored separately from individual launchers
    - the individual launchers (*.sht) are processed through the C pre-processor to get the final launchers (*.sh); there is a "launcher-scripts" target in the Makefile for this
- there is a new abstract class, script_based_launcher, which is used by all the other (current) launchers (single, multi, mpi)

Proper testing needs to be added for all the knobs, but, for now, I'm committing the infrastructure and only doing some basic local tests.


src/psi/j/executors/local.py
```diff
@@ -205,14 +205,17 @@ def submit(self, job: Job) -> None:
...
 
         p = _ChildProcessEntry(job, self)
 
         try:
             with job._status_cv:
                 if job.status.state == JobState.CANCELED:
                     raise SubmitException('Job canceled')
+            if logger.isEnabledFor(logging.DEBUG):
+                logger.debug('Running {},  out={}, err={}'.format(args, spec.stdout_path,
+                                                                  spec.stderr_path))
```
andre-merzky, line 218: The test for `logger.isEnabledFor('DEBUG')` is already performed by the implementation of `logger.debug` - so why are the logging calls shielded?  I am probably missing something...

If it is to avoid early string expansion on the `format`, that is usually addressed by:
```py
logger.debug('Running %s,  out=%s, err=%s', args, spec.stdout_path, spec.stderr_path)
```
the logging module will the only expand if the respective log level is enabled.  There is probably an option for `format` like string expansion as you seem to prefer that, but I don't actually know...

src/psi/j/launchers/script_based_launcher.py
```diff
@@ -0,0 +1,83 @@
...
+        - the pre- and post- launcher scripts, or empty strings if they are not specified
+
+    Additional positional arguments to the script can be specified by subclasses by overriding
+    the :func:`~psi.j.launchers.script_based_launcher.ScriptBasedLauncher._get_additional_args`
+    function.
+
+    The remaining arguments to the script are the job executable and arguments.
+
+    A simple script library is provided in scripts/lib.sh. Its use is optional and it is intended
+    to be included in a main launcher script using a standard C preprocessor. It does the following:
```
andre-merzky, line 34: I am a bit confused why the C preprocessor is needed - couldn't the same be achieved by this just being a shell script?  But I am probably missing something again...  You seem to invoke the preprocessor to build the actual scripts  during install - but you could also just install `lib.sh` and source it at runtime?  Or are you trying to optimize the sourcing away?

andre-merzky (approved):
I have some comments (more questions really), but nothing suspicious sticks out to me.

src/psi/j/executors/local.py
```diff
@@ -205,14 +205,17 @@ def submit(self, job: Job) -> None:
             raise InvalidJobException('Missing specification')
 
         launcher = self._get_launcher(self._get_launcher_name(spec))
-        args = launcher.get_launch_command(spec)
+        args = launcher.get_launch_command(job)
```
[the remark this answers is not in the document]
hategan: The job id. I know, that guy!

src/psi/j/executors/local.py
hategan: Yes, but the check in logger.debug comes after the message string is interpolated.

hategan: That said, I should probably use the built-in formatting, as you suggest.

src/psi/j/launchers/script_based_launcher.py
hategan: I'm trying to optimize the sourcing away. This is probably premature optimization.

I am also trying to avoid figuring out where lib.sh is when the parent script is invoked.

There are probably alternative ways to do things here which may be better.

andre-merzky: > I'm trying to optimize the sourcing away. This is probably premature optimization.

Indeed, I would say it is... :-)

> I am also trying to avoid figuring out where lib.sh is when the parent script is invoked. 
> There are probably alternative ways to do things here which may be better.

Fair point.  Just adding this as input:

we install shell scripts along with the python sources, so the python module always finds the script with 
```py
script = '%s/script.sh` % os.path.dirname(__file__)
```

Probably not foolproof, but it served us well so far.  Another option is to install into `bin`: `.` and `source` [actually search the path](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_18), so it will be found via `$PATH`.

hategan: > ```python
> script = '%s/script.sh` % os.path.dirname(__file__)
> ```

That's what's happening. But there's also the step of sourcing lib.sh from another .sh, and there is no __file__ at that stage. It's possible to figure it out using $0, but I opted against that for now. 

There's also the soon-to-be-an-issue of submit scripts (e.g., for PBS) and whether launcher scripts would be called from them or inlined into them.

src/psi/j/executors/local.py
hategan: Removed the guards in the local executor. Will also remove in other places, but not in this PR.

hategan:
@SteVwonder, would you like to take a look at this?

SteVwonder (commented):
I haven't given it a full trial run, but a review of the code generally LGTM.  Seems as though there are some test issues though:

```
❯ make checks                                                                                                                                                           23:02:06 ()
mypy --config-file=.mypy --strict src tests
Success: no issues found in 22 source files
flake8 src
flake8 tests
tests/test_local_jobs.py:74:1: W391 blank line at end of file
make: *** [stylecheck] Error 1

❯ make tests                                                                                                                                                            23:02:55 ()
PYTHONPATH=/Users/herbein1/Repositories/exaworks/psi-j-python/src python -m pytest
=============================================================================== test session starts ================================================================================
platform darwin -- Python 3.9.4, pytest-6.2.3, py-1.10.0, pluggy-0.13.1
rootdir: /Users/herbein1/Repositories/exaworks/psi-j-python
collected 8 items                                                                                                                                                                  

tests/test_doc_examples.py ..                                                                                                                                                [ 25%]
tests/test_local_jobs.py .F.F..                                                                                                                                              [100%]

===================================================================================== FAILURES =====================================================================================
_____________________________________________________________________________ test_simple_job_redirect _____________________________________________________________________________

    def test_simple_job_redirect() -> None:
        with TemporaryDirectory() as td:
            outp = Path(td, 'stdout.txt')
            job = Job(JobSpec(executable='/bin/echo', arguments=['-n', '_x_'], stdout_path=outp))
            exec = JobExecutor.get_instance('local')
            exec.submit(job)
            job.wait()
            f = outp.open("r")
            contents = f.read()
>           assert contents == '_x_'
E           AssertionError: assert '' == '_x_'
E             - _x_

tests/test_local_jobs.py:23: AssertionError
___________________________________________________________________________________ test_cancel ____________________________________________________________________________________

    def test_cancel() -> None:
        job = Job(JobSpec(executable='/bin/sleep', arguments=['1']))
        exec = JobExecutor.get_instance('local')
        exec.submit(job)
        job.wait(target_states=[JobState.ACTIVE])
        job.cancel()
        status = job.wait()
        assert status is not None
>       assert status.state == JobState.CANCELED
E       assert FAILED == CANCELED
E         Use -v to get the full diff

tests/test_local_jobs.py:47: AssertionError
============================================================================= short test summary info ==============================================================================
FAILED tests/test_local_jobs.py::test_simple_job_redirect - AssertionError: assert '' == '_x_'
FAILED tests/test_local_jobs.py::test_cancel - assert FAILED == CANCELED
=========================================================================== 2 failed, 6 passed in 1.63s ============================================================================
make: *** [tests] Error 1
```

src/psi/j/launchers/scripts/lib.sh
```diff
@@ -0,0 +1,42 @@
+#define _HASH_ #
+#define _BIN_BASH_ #!/bin/bash
+
```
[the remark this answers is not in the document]
hategan: They are launched with bash explicitly. That said, _BIN_BASH_ is only useful at the beginning of lib.sh. Unfortunately, it does not work since (at least gcc) the preprocessor seems to insert a space before the #!/bin/bash. So this whole #include needs some work. But not right now.

hategan:
> there are some test issues though:

Interesting. I am not getting these. Unfortunately, I do not have a mac close by, so I'm not sure how to approach the issue.

SteVwonder (approved):
LGTM! And we will spin out the Mac issue into a separate issue.

PR #13, opened 2021-04-02
3.6 compatibility

hategan:
The problem was caused by "from __future__ import annotations", which is there to support forward type definitions (e.g., `class A: def fn(a: A):...` instead of having to quote A - `class A: def fn(a: 'A'):...`)

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -25,8 +25,6 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
-        python -m pip install flake8 pytest
         pip install -r requirements.txt
         pip install -r requirements-dev.txt
     - name: Lint with flake8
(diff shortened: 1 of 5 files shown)


andre-merzky (approved):
LGTM

SteVwonder (approved):
LGTM!

SteVwonder:
Now that #6 is in, do we want to squash and merge this?  Or rebase it?  Not sure how much we care about the duplicate commits.

hategan:
> Now that #6 is in, do we want to squash and merge this? Or rebase it? Not sure how much we care about the duplicate commits.

This was rebased on #6, so should be fine now that #6 has been merged.

hategan:
@andre-merzky, @SteVwonder, can I merge this?

andre-merzky:
From my end yes, and happy to push the button.  Please let me use that opportunity for a question though: @SteVwonder: should the merge commit be squashed into a single commit, or is that up to the owner of the PR?

SteVwonder:
> @andre-merzky, @SteVwonder, can I merge this?

Yeah. I was initially confused by the Github diff, since it appears to show this PR adding the entire contents of the workflow yaml.  I pulled down the branch and `git diff upstream/master` and that made way more sense.  Sorry, not used to reviewing PRs that have merge commits in them.

> Please let me use that opportunity for a question though: @SteVwonder: should the merge commit be squashed into a single commit, or is that up to the owner of the PR?

I don't think we have that documented anywhere.  In a rebase-based fork/pull or C4 gitflow, I think you would normally leave all the commits since they will apply cleanly and you keep the more precise history.  Idk what is typically done in gitflow, for example.  I know Spack makes extensive use of "squash and merge".  What are you two more accustomed to?

For this PR, I think it may make more sense to do a "squash and merge".  I'll leave that choice up to @hategan though.  Any preference @hategan?

hategan:
> I know Spack makes extensive use of "squash and merge". What are you two more accustomed to?

I try to keep my commits small on purpose, so squashing would seem silly to me, but maybe I'm missing something.

andre-merzky:
>  Idk what is typically done in gitflow, for example.

We usually leave all commits, for two reasons: (a) we don't care about clean history :-P, (b) it makes `git bisect` more efficient.

Thanks both!

PR #23, opened 2021-05-06
Fixed a nested path executor loading issue

hategan:
Fixed an error in which nested paths in PYTHONPATH can lead to the same executor being loaded twice, which can lead to an executor loading error


andre-merzky (approved):
I guess that was fun to track down :-P
LGTM!

hategan:
> I guess that was fun to track down :-P

I've had worse :)

SteVwonder (approved):

SteVwonder:
Since this is double approved, going to merge this in.

Tested locally and it works for me.

PR #26, opened 2021-05-12
Saga connector

hategan:
A prototype SAGA connector. There are a few issues in the code, marked by `# TODO:`. We should consider fixing those before merging.

### DIFF
--- a/src/psi/j/executors/saga.py
+++ b/src/psi/j/executors/saga.py
@@ -1,0 +1,222 @@
+import atexit
+import logging
+from threading import RLock
+from typing import Optional, Dict, List
+
+import radical.saga as rs
+import radical.saga.job.constants as ct
+import time
+
+from psi.j import JobExecutor, JobExecutorConfig, JobState, JobStatus, Job, \
+    SubmitException, InvalidJobException
+
+logger = logging.getLogger(__name__)
+
+
+# map SAGA states to jpsi states.
+_STATE_MAP = {
+    rs.job.NEW: JobState.NEW,
+    rs.job.PENDING: JobState.QUEUED,
+    rs.job.RUNNING: JobState.ACTIVE,
+    rs.job.DONE: JobState.COMPLETED,
+    rs.job.FAILED: JobState.FAILED,
+    rs.job.CANCELED: JobState.CANCELED,
+}
+
+
+class _JobMapping:
+    def __init__(self, jpsi_job: Job, saga_job: Optional[rs.job.Job] = None):
+        self.jpsi_job = jpsi_job
+        self.saga_job = saga_job
+
+
+class SagaExecutor(JobExecutor):
+    """
+    An executor that use SAGA to run jobs.
+
+    For details about SAGA, see https://en.wikipedia.org/wiki/Simple_API_for_Grid_Applications
+    and related links.
+    """
+
+    assert('job' in dir(rs)), 'incomplete installation'
+
+    def __init__(self, url: Optional[str] = None,
+                 config: Optional[JobExecutorConfig] = None) -> None:
+        """
+        Initializes a `SagaExecutor`.
+
+        Parameters
+        ----------
+        url
+            An URL representing a SAGA resource manager to connect to.
+        config
+            Not used.
+        """
+        super().__init__(url=url, config=config)
+
+        self._jobs = dict()  # type: Dict[str, _JobMapping]
+        self._lock = RLock()
+        self._js = rs.job.Service(self.url)
+        self._cb = None
+        atexit.register(self._close)
+
+    def _close(self) -> None:
+        if self._js:
+            self._js.close()
+
+    def __del__(self) -> None:
+        """
+        Removes process shutdown hooks added earlier.
+
+        The hooks are used to close the SAGA service used by this executor.
+        """
+        atexit.unregister(self._close)
+        self._close()
+
+    def _state_cb(self, saga_job: rs.job.Job, metric: object, ctx: object) -> bool:
+        jpsi_uid = saga_job._jex_uid
+        jpsi_job = self._get_jpsi_job(jpsi_uid)
+        ec = None
+
+        if saga_job.state in [rs.FAILED, rs.DONE]:
+            ec = saga_job.exit_code
+
+        logger.debug('%s --> %s', jpsi_uid, saga_job.state)
+
+        job_status = JobStatus(_STATE_MAP[saga_job.state], time.time(), exit_code=ec)
+        jpsi_job._set_status(job_status, self)
+        if self._cb:
+            self._cb.job_status_changed(jpsi_job, job_status)
+
+        return True
+
+    def submit(self, job: Job) -> None:
+        """
+        Submits the specified :class:`~psi.j.Job` to be run using SAGA.
+
+        Parameters
+        ----------
+        job
+            The :class:`~psi.j.Job` to be submitted.
+        """
+        # derive SAGA job description and submit it
+        job_mapping = _JobMapping(job)
+        with self._lock:
+            self._jobs[job.id] = job_mapping
+        job._executor = self
+
+        jd = self._job_2_descr(job)
+        saga_job = self._js().create_job(jd)
+        saga_job._jex_uid = job.id
+
+        job_mapping.saga_job = saga_job
+        saga_job.add_callback(rs.STATE, self._state_cb)
+
+        # TODO: the backend attribute is not standard; why do we need it?
+        self._update_job_status(job, JobStatus(JobState.QUEUED, metadata={'backend': self.url}))
+
+        self._update_job_status(job, JobStatus(JobState.ACTIVE))
+
+        saga_job.run()
+
+    def _get_saga_job(self, uid: str) -> rs.job.Job:
+        with self._lock:
+            return self._jobs[uid].saga_job
+
+    def _get_jpsi_job(self, uid: str) -> Job:
+        with self._lock:
+            return self._jobs[uid].jpsi_job
+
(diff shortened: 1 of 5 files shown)


src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+        saga_job = self._js().create_job(jd)
+        saga_job._jex_uid = job.id
+
+        job_mapping.saga_job = saga_job
+        saga_job.add_callback(rs.STATE, self._state_cb)
+
+        # TODO: the backend attribute is not standard; why do we need it?
+        self._update_job_status(job, JobStatus(JobState.QUEUED, metadata={'backend': self.url}))
+
+        self._update_job_status(job, JobStatus(JobState.ACTIVE))
```
SteVwonder, line 118: I'm curious as to why the job immediately gets moved to `ACTIVE`.  Shouldn't that be handled by the state_cb since the job could be stuck in the queue?  Maybe this is just a behavior of SAGA that I'm unaware of.

src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+
+class SagaExecutor(JobExecutor):
+    """
+    An executor that use SAGA to run jobs.
+
+    For details about SAGA, see https://en.wikipedia.org/wiki/Simple_API_for_Grid_Applications
+    and related links.
+    """
+
+    assert('job' in dir(rs)), 'incomplete installation'
```
andre-merzky, line 41: How can that happen?

src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+
+        self._jobs = dict()  # type: Dict[str, _JobMapping]
+        self._lock = RLock()
+        self._js = rs.job.Service(self.url)
+        self._cb = None
+        atexit.register(self._close)
+
+    def _close(self) -> None:
+        if self._js:
+            self._js.close()
```
andre-merzky, line 65: add `self._js = None` to avoid repeated close calls?  Otherwise the `if` will always be `True`

src/psi/j/executors/saga.py
andre-merzky: I agree with that observation - the SAGA's `RUNNING` state should translate to `ACTIVE` (as also is encoded in the `_STATE_MAP`).

src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+            The :class:`~psi.j.Job` to be submitted.
+        """
+        # derive SAGA job description and submit it
+        job_mapping = _JobMapping(job)
+        with self._lock:
+            self._jobs[job.id] = job_mapping
+        job._executor = self
+
+        jd = self._job_2_descr(job)
+        saga_job = self._js().create_job(jd)
```
andre-merzky, line 109: I am confused about the `_js()` (instead of `_js`) - did the SAGA job service convert into a `weakref` somewhere?  Why only here?  I may be missing something...

src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+        Parameters
+        ----------
+        job
+            The job to attach.
+        native_id
+            A PSI-J job ID
+
+        """
+        # This probably needs to be re-worked. I would assume that the meaning of attach()
+        # here is not to attach to a saga job, but to attach to a scheduler job
```
andre-merzky, line 159: I'm afraid I don't understand that comment.  What would a `scheduler job` be in this context?  Like a Slurm job?  How would one attach to that w/o going via SAGA (and thus the SAGA job)?

src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # try to find job in known jobs
+        with self._lock:
+            for job_mapping in self._jobs.values():
+                try:
+                    if job_mapping.jpsi_job.status.metadata:
+                        job_native_id = job_mapping.jpsi_job.status.metadata['native_id']
+                        if job_native_id == native_id:
+                            # TODO: what is the meaning here?
```
andre-merzky, line 168: What is the meaning of 'here'? :-)
The commented line below looks correct to me.

src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+        # otherwise re-connect from backend:
+        saga_job = None
+        try:
+            saga_job = self._js.get(native_id)
+        except Exception:
+            pass
+
+        if not saga_job:
+            # TODO: There is a discussion on the suitability of cancel() throwing a SubmitException.
+            # TODO: that same discussion would apply here
```
andre-merzky, line 183: Indeed, but that could possibly be out of scope for this PR?

andre-merzky (commented):
Thanks for the PR - I left a couple of questions /  comments.

src/psi/j/executors/saga.py
hategan: That was there in the original version. I tried to make minimal changes.

So if you think it should be removed, it should probably be removed.

src/psi/j/executors/saga.py
hategan: True. Either the check is redundant because other things ensure that this is only called once, or it is not, but then what you say.

src/psi/j/executors/saga.py
hategan: No weakrefs. I think it's a mistake that should be fixed.

src/psi/j/executors/saga.py
hategan: It doesn't look like one can attach to a job that was submitted in another process, since it relies on `_jobs`, which is only updated by `submit()`.

src/psi/j/executors/saga.py
hategan: It refers to the commented-out return statement. Why was there one in the original? Maybe an artifact from when the API wasn't quite settled?

src/psi/j/executors/saga.py
hategan: Indeed out of scope, but I think it's a remarkable use case, so I thought I'd mark it.

src/psi/j/executors/saga.py
andre-merzky: Ha! :-)  I think this can be removed by now.

src/psi/j/executors/saga.py
andre-merzky: Ack, probably.

src/psi/j/executors/saga.py
```diff
@@ -0,0 +1,222 @@
...
+            raise SubmitException('Unknown native id: "{}'.format(native_id))
+
+        # we found the job - recreate state
+        state = _STATE_MAP[saga_job.state]
+        if state.final:
+            self._update_job_status(job, JobStatus(state, exit_code=saga_job.exit_code))
+        else:
+            self._update_job_status(job, JobStatus(state))
```
andre-merzky, line 191: before returning, we probably want to update the job mapping in `self._jobs` so that we can update the jpsi_job when the saga_job gets any state notifications?

andre-merzky:
@hategan : I pushed a small iteration to the saga connector.  It seems to work now, in the sense that jobs get started and collected.  I did not work on the various comments from above.

One problem I stumbled over: I don't see the class destructor being called on termination.  The saga session needs an explicit `close()` for clean shutdown (and otherwise termination will hang on a thread lock).  Any insight on why the d'tor is not called, and what to do about it?

hategan:
> One problem I stumbled over: I don't see the class destructor being called on termination.

There is a shutdown handler (using atexit) called at termination which calls close() on the job service. Is that not working?

andre-merzky:
> > One problem I stumbled over: I don't see the class destructor being called on termination.
> 
> There is a shutdown handler (using atexit) called at termination which calls close() on the job service. Is that not working?

Nope, it does not - the interpreter does not seem to go into exit on it's own.

hategan:
> Nope, it does not - the interpreter does not seem to go into exit on it's own.

This would imply that the atexit module is broken. While possible, it seems unlikely. So maybe let's try to troubleshoot this and see what's happening.

hategan:
> > Nope, it does not - the interpreter does not seem to go into exit on it's own.
> 
> This would imply that the atexit module is broken. While possible, it seems unlikely. So maybe let's try to troubleshoot this and see what's happening.

We clarified on a call that the issue is not that the atexit handler fails to be called, but that the process does not exit at all.

In this instance, there are two threads left alive after the main thread exists:
1. a shell_job._job_state_monitor
2. a threading.Timer started by ShellJobService.initialize

The former is a daemon thread. The timer is not.

The quickest solution would probably be to add, in ShellJobService.initialize:
```diff
self._ping = mt.Timer(_PING_DELAY, self._ping_cb)
+ self._ping.daemon = True
self._ping.start()
```

Then base the adaptor on an updated version of SAGA.

A hackish alternative would be to add, in the saga adapter:
```python
def _patched_initialize(self):
    self._old_initialize()
    self._ping.cancel()
    self._ping = mt.Timer(shell_job._PING_DELAY, self._ping_cb)
    self._ping.daemon = True
    self._ping.start()

ShellJobService._old_initialize = ShellJobService.initialize
ShellJobService.initialize = _patched_initialize
```

This comment is not meant to rule out the possibility of adding a `JobExecutor.close()` method in the spec, but to explore possible solutions.

andre-merzky:
@hategan : please press the merge button, if you don't mind...

hategan:
> @hategan : please press the merge button, if you don't mind...

A conflict popped up in the tests. I'll need to fix that.

PR #28, opened 2021-05-14
Properly handle launcher errors

### LINKED ISSUE #25: Mac support
Local connector tests are failing on Mac.  @hategan is investigating debuggin on AWS

hategan:
Commit 47ce6db describes the general problem in more detail.

In addition, this PR removes comments in launcher script templates. The templates are processed with the C pre-processor. On linux with somewhat recent versions of GCC, the default behavior is to remove comments from the output. On OS X, the default `cpp` seems to be the LLVM clang compiler, which keeps comments after pre-processing. These comments, however, are not valid comments in a script. So this should also fix #25.

### DIFF
--- a/src/psi/j/executors/local.py
+++ b/src/psi/j/executors/local.py
@@ -6,12 +6,11 @@ import threading
 import time
 from abc import ABC, abstractmethod
 from distutils.version import StrictVersion
-from pathlib import Path
-from typing import IO, Union, Any, Optional, Dict, List, Type
+from typing import Optional, Dict, List, Type, Tuple
 
 import psutil
 
-from psi.j import InvalidJobException, SubmitException
+from psi.j import InvalidJobException, SubmitException, Launcher
 from psi.j import Job, JobSpec, JobExecutorConfig, JobState, JobStatus
 from psi.j import JobExecutor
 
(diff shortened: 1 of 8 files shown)


andre-merzky (approved):
LGTM

src/psi/j/executors/local.py
```diff
@@ -245,16 +228,23 @@ def cancel(self, job: Job) -> None:
...
     def _process_done(self, p: _ProcessEntry) -> None:
-        p.close()
         assert p.exit_code is not None
+        message = None
         if p.exit_code == 0:
             state = JobState.COMPLETED
         elif p.exit_code < 0 and p.kill_flag:
             state = JobState.CANCELED
         else:
+            # We want to capture errors in the launcher scripts. Since, under nomral circumstances,
```
SteVwonder, line 238:
```suggestion
            # We want to capture errors in the launcher scripts. Since, under normal circumstances,
```

SteVwonder (approved):
LGTM! Happy to merge as-is, but I noticed a small typo in a comment.  Will hold off in case you want to fix (if so, feel free to squash the last two commits since the one is a fixup of the other).  Otherwise, let us know and I'm happy to push the button now.

PR #43, opened 2021-07-14
Fix `version` accessor

### LINKED ISSUE #41: Executors missing `__VERSION__`
```
❯ python3 -c 'from psi.j.job_executor import JobExecutor; exec = JobExecutor.get_instance("local"); print(exec.version)'                                                                                   19:35:22 ()
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/herbein1/Repositories/exaworks/psi-j-python/.virtualenv/lib/python3.9/site-packages/psi/j/job_executor.py", line 69, in version
    return cast(Version, getattr(self.__class__, '__VERSION__'))
AttributeError: type object 'LocalJobExecutor' has no attribute '__VERSION__'
```

SteVwonder:
Fix executors' `version` property getter.

Oh, and I snuck a gitignore change in since `make` creates a `build` directory that git wants to track by default.

Closes #41

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ psi_j_python.egg-info/
 
 venv*
 .venv*
+build/
--- a/src/psi/j/job_executor.py
+++ b/src/psi/j/job_executor.py
@@ -66,7 +66,7 @@ class JobExecutor(ABC):
     @property
     def version(self) -> Version:
         """Returns the version of this executor."""
-        return cast(Version, getattr(self.__class__, '__VERSION__'))
+        return cast(Version, getattr(self.__class__, '_VERSION_'))
 
     @abstractmethod
     def submit(self, job: Job) -> None:
--- a/tests/test_executor_versions.py
+++ b/tests/test_executor_versions.py
@@ -1,0 +1,12 @@
+#!/usr/bin/env python3
+
+# This is meant as a simple test file to check if psi/j was installed successfully
+
+from distutils.version import Version
+
+from psi.j import JobExecutor
+
+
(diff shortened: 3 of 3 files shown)


hategan:
> I initially just wanted to handle #41, but that was difficult locally due to the requirement to have `flux` installed and importable. So this PR first removes that requirement (which we briefly discussed on a previous  call. Then it fixes the `version` accessor and adds a test.

I think removing the requirement for Flux to be installed and importable is fine. But I worry about the error being silently ignored. Both legitimate errors in importing flux classes as well as flux not being installed are presented to the user as the absence of the flux executor. This could potentially send users on very wasteful wild goose chases.

I would suggest either moving executors with heavy dependencies to their own, separately installable packages, as I think we might have discussed as a possibility in the past. Alternatively, save the import exception and throw it when the user attempts to instantiate the flux executor. I think the latter can be made as a generic solution in the mechanism used to discover/instantiate executors.

> 
> I'm happy to split these two changes into two separate PRs if desired since we technically don't have an issue open on the first change. Oh, and I snuck a gitignore change in since `make` creates a `build` directory that git wants to track by default.

This is fine with me.

SteVwonder:
ACK. All valid points.  I stripped this PR back to just the `version` fix, and I'll open an issue to track the handling of executors with dependencies.

hategan (approved):

SteVwonder:
Rebased and mypy/flake8 were made happy.  I think this is ready for @hategan to re-review and/or merge.

SteVwonder:
Thanks!

hategan:
> Thanks!

No, thank you!

PR #55, opened 2021-10-21
Remote testing

hategan:
This PR adds remote test capabilities. There are a few components:

- test output/log capture, which happens in conftest.py
- save/uploads to a testing server, also in conftest.py if requested
- a "ci-runner" (for lack of a better name), which, using testing.conf can clone the git repo, run tests on some branches, and upload the results to said testing server

There is a corresponding server in the psi-j-testing-server repo. It is currently deployed on GCE at testing.exaworks.org, which, incidentally, can also be used to visualize test results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,10 @@ verbose-tests: launcher-scripts
 	PYTHONPATH=$(CWD)/src:${PYTHONPATH} ${PYTHON} -m pytest -v --log-format="%(asctime)s %(levelname)s %(message)s" \
 		           --log-date-format="%Y-%m-%d %H:%M:%S" --log-cli-level=DEBUG
 
+.PHONY: ci-tests
+ci-tests:
+	${PYTHON} tests/ci_runner.py
+
 
 .PHONY: typecheck
 typecheck:
--- a/Makefile
+++ b/Makefile
@@ -10,11 +10,11 @@ CWD = $(shell pwd)
 PYTHON = $(shell if python --version 2>&1 | egrep -q 'Python 3\..*' ; then echo "python"; else echo "python3"; fi)
 
 .PHONY: tests
-tests:
+tests: launcher-scripts
 	PYTHONPATH=$(CWD)/src:${PYTHONPATH} ${PYTHON} -m pytest -v $(TESTARGS)
 
 .PHONY: verbose-tests
-verbose-tests:
+verbose-tests: launcher-scripts
 	PYTHONPATH=$(CWD)/src:${PYTHONPATH} ${PYTHON} -m pytest -v --log-format="%(asctime)s %(levelname)s %(message)s" \
 		           --log-date-format="%Y-%m-%d %H:%M:%S" --log-cli-level=DEBUG
 
--- a/psij-ci-run
+++ b/psij-ci-run
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-python tests/ci_runner.py
+if python --version 2>&1 | egrep -q 'Python 3\..*' >/dev/null 2>&1 ; then
+    PYTHON="python"
+else
+    PYTHON="python3"
+fi
+
+$PYTHON tests/ci_runner.py
--- a/psij-ci-run
+++ b/psij-ci-run
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+python tests/ci_runner.py
--- a/psij-ci-run
+++ b/psij-ci-run
@@ -1,5 +1,49 @@
 #!/bin/bash
 
+usage() {
+    cat <<EOF
+Usage: ./psij-ci-run [--help] [--with-conda <conda_env_name>] 
+            [--with-venv <venv_name>]
+    --help          Displays this message
+    --with-conda    Specifies a conda environment to activate before running 
+                    the tests
+    --with-venv     Specifies a virtualenv environment to load before running
+                    the tests
+EOF
+}
+
+failifempty() {
+    if [ "$2" == "" ]; then
+        echo "Missing parameter for $1"
+        usage
+        exit 1
+    fi
+}
+
+while [ "$1" != "" ]; do
+    case "$1" in
+        --with-conda)
+            failifempty "$1" "$2"
+            conda activate "$2"
+            shift 2
+            ;;
+        --with-venv)
+            failifempty "$1" "$2"
+            source "$2/bin/activate"
+            shift 2
+            ;;
+        --help)
+            usage
+            exit 0
+            ;;
+        *)
+            echo "Unrecognized command line option: $1."
+            usage
+            exit 1
+    esac
+done
+
+
 if python --version 2>&1 | egrep -q 'Python 3\..*' >/dev/null 2>&1 ; then
     PYTHON="python"
 else
--- a/psij-ci-setup
+++ b/psij-ci-setup
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+FORCE=0
+
+if [ "$1" == "-f" ]; then
+    FORCE=1
+    shift
+fi
+
+MYPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+echo
+echo "================================================================"
+echo "This script will install requirements for the PSI/J CI tests and"
+echo "add a cron job to run the tests once a day at some random time. "
+echo "================================================================"
+echo
+
+RESPONSE=""
+
+while [ "$RESPONSE" != "C" ] && [ "$RESPONSE" != "X" ]; do
+
+    echo -n "Would you like to (C)ontinue or E(x)it? "
+    read -n1 RESPONSE
+    echo
+    RESPONSE=${RESPONSE^}
+    
+    if [ "$RESPONSE" == "X" ]; then
+        echo "Operation canceled"
+        exit 1
+    fi
+done
+
+cd "$MYPATH"
+
+pip install --user -r requirements-dev.txt
+
+
+HOUR=`echo $(($RANDOM % 24))`
+MINUTE=`echo $(($RANDOM % 60))`
+MYPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+if crontab -l 2>/dev/null | grep "psij-ci-run" >/dev/null && [ "$FORCE" != "1" ]; then
+    EXISTING=`crontab -l 2>/dev/null | grep "psij-ci-run"`
+    echo
+    echo "================================================================"
+    echo "Error: a crontab for PSI/J tests already exists: "
+    echo ">>> $EXISTING"
+    echo
+    echo "You can edit your crontab with \"crontab -e\" and remove the    "
+    echo "existing entry, then re-run this tool. If you are certain that  "
+    echo "you want to install multiple entries, you can re-run this script"
+    echo "with the \"-f\" flag.                                           "
+    echo "================================================================"
+    exit 2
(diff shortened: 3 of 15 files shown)


hategan:
> Finally done adding commits, eh? ;)

Maaaaybe?

psij-ci-setup
```diff
@@ -0,0 +1,66 @@
...
+RESPONSE=""
+
+while [ "$RESPONSE" != "C" ] && [ "$RESPONSE" != "X" ]; do
+
+    echo -n "Would you like to (C)ontinue or E(x)it? "
+    read -n1 RESPONSE
+    echo
+    RESPONSE=${RESPONSE^}
+    
+    if [ "$RESPONSE" == "X" ]; then
```
andre-merzky, line 28: logic should be reversed - like, break when the user enters `?`.  BTW: this prompts for `E(x)it` (lowercase `x`) but checks for uppercase `X`.

psij-ci-setup
```diff
@@ -0,0 +1,66 @@
...
+        echo "Operation canceled"
+        exit 1
+    fi
+done
+
+cd "$MYPATH"
+
+pip install --user -r requirements-dev.txt
```
andre-merzky, line 36: Installing in `--user` is not a good idea for shared file systems which are used on resources with different architectures.  Have been there, done that, and lost a week debugging why the heck some module did not work... :-P  Also, that behavior would be surprising if the script runs in a virtual or conda env I think?

Makefile
```diff
@@ -10,14 +10,18 @@ CWD = $(shell pwd)
 PYTHON = $(shell if python --version 2>&1 | egrep -q 'Python 3\..*' ; then echo "python"; else echo "python3"; fi)
```
andre-merzky, line 10: why not always `python3`?

psij-ci-run
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python tests/ci_runner.py
```
andre-merzky, line 3: this should be `python3`.  On my system:
```sh
$ python -V
Python 2.7.17
```

tests/test_executor.py
```diff
@@ -3,86 +3,109 @@
...
 
-@pytest.mark.parametrize('name,url', tests)
-def test_attach(name: str, url: Optional[str]) -> None:
+def test_attach(execparams: ExecutorTestParams) -> None:
     job = Job(JobSpec(executable='/bin/sleep', arguments=['1']))
-    jex = JobExecutor.get_instance(name=name, url=url)
-    jex.submit(job)
-    job.wait(target_states=[JobState.ACTIVE])
+    exec = _get_executor_instance(execparams, job)
+    exec.submit(job)
```
andre-merzky, line 53: `exec`  is a keyword (applies elsewhere too)

andre-merzky (commented):
Some minor comments on the code, will run some tests now :-)

andre-merzky:
Test result: installing in a VE is straight forward.  The result seems to get pushed, there is a new entry in the table (`rivendell`, my local machine).  Some tests failed actually (`No such launcher: "mpi"`), but the entry is grey (I would have expected red?).  Not sure how to interpret that.

When the runner starts, I see this message
```sh
$ python3 tests/ci_runner.py 
Reading configuration... OK
Cloning repository... Cloning into 'code'...
remote: Enumerating objects: 1383, done.
remote: Counting objects: 100% (1383/1383), done.
remote: Compressing objects: 100% (701/701), done.
remote: Total 1383 (delta 764), reused 1132 (delta 577), pack-reused 0
Receiving objects: 100% (1383/1383), 225.39 KiB | 1.39 MiB/s, done.
Resolving deltas: 100% (764/764), done.
OK
Branch 'cobalt-executor' set up to track remote branch 'cobalt-executor' from 'origin'.
Switched to a new branch 'cobalt-executor'
Testing branch "cobalt-executor"... running launcher_scripts
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --upload-results --run-id --id "rivendell" --branch-name-override cobalt-executor --maintainer-email andre@merzky.net --server-url https://testing.exaworks.org --key "rivendell" --max-age 48
  inifile: None
  rootdir: /tmp/tmphgvajnlp/code

OK
Branch 'testing' set up to track remote branch 'testing' from 'origin'.
```

I interrupted the run a couple of time as I thought with was an error (it is UPPERCASED after all! ;-) ), but it turned out to be non-fatal.  Probably should be filtered out?

psij-ci-setup
hategan: I'm assuming that the main scenario where we would want this to function is resources that the user does not have root access to. I'd be happy to hear alternatives, but system installs don't seem to be in the realm of plausible here.

Makefile
hategan: Can I assume that all systems have Python 3 as python3? I believe I've encountered cases where this was no true, but I'm not entirely sure.

psij-ci-run
hategan: Ah, thanks! Yes, this should follow the same rules as above.

tests/test_executor.py
hategan: Thank you. Will update.

psij-ci-setup
hategan: > logic should be reversed - like, break when the user enters `?`.

Good point.

> BTW: this prompts for `E(x)it` (lowercase `x`) but checks for uppercase `X`.

The input is converted to capital letters in line 26.

hategan:
> but the entry is grey (I would have expected red?). Not sure how to interpret that.

You live in the future, so your tests haven't happened yet!

It might be because the server or interface use local time to build the table. I will need to make sure that all times are in UTC.

andre-merzky:
> You live in the future, so your tests haven't happened yet!

Indeed, my clone agrees! :-D

FWIW, deploying in conda worked just as well, it resulted in the same output.

hategan:
> FWIW, deploying in conda worked just as well, it resulted in the same output.

Interesting, but is it maybe because the dependencies were already installed?

If it's not too much work, can you wipe pystache from the system and try to install it under conda with pip?

psij-ci-setup
hategan: > > logic should be reversed - like, break when the user enters `?`.
> 
> Good point.
> 
> >

Actually wait. The behavior, as it is, because of the while loop, is to keep asking if neither c or x are entered. So I think this works as intended.

psij-ci-run
hategan: Is it the same with pip? Aka, do I need to say `pip3` on your system to do the right thing?

Makefile
andre-merzky: Oh, I agree with that - but I would expect an install in a virtualenv or conda env to be preferred.  At the end it is hard to map the gazillion of python deployment options into a script :-/

psij-ci-run
andre-merzky: Yes, same with pip indeed.

andre-merzky:
> Interesting, but is it maybe because the dependencies were already installed?
> 
> If it's not too much work, can you wipe pystache from the system and try to install it under conda with pip?

You are right, that failed in a new env.  This fixes it:

```sh
pip install "setuptools<58"
```

Ultimately that needs fixing in `pystache` I'm afraid...

tests/test_executor.py
```diff
@@ -3,86 +3,109 @@
...
     assert status.exit_code is not None
     assert status.exit_code != 0
 
 
-@pytest.mark.parametrize('name,url', tests)
-def test_missing_executable(name: str, url: Optional[str]) -> None:
+def test_missing_executable(execparams: ExecutorTestParams) -> None:
     job = Job(JobSpec(executable='/bin/no_such_file_or_directory'))
-    jex = JobExecutor.get_instance(name=name, url=url)
+    exec = _get_executor_instance(execparams, job)
```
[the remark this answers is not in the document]
jameshcorbett: Fixed by d7d88e9, marking as resolved.

hategan:
> Ultimately that needs fixing in `pystache` I'm afraid...

We should figure out why pip installs it from source rather than the wheel, since it might provide us with the easiest solution.

tests/test_executor.py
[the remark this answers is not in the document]
hategan: Ooops! Sorry.

hategan:
> > Ultimately that needs fixing in `pystache` I'm afraid...
> 
> We should figure out why pip installs it from source rather than the wheel, since it might provide us with the easiest solution.

More on this. The wheel for pystache 0.5.4 was removed from pypi and, unless you have a cached one locally or if you run with --no-cache, pip will use the source package which fails as described above.

There is a promising fork of pystache whose maintainer is attempting to take over the pypi package (see https://github.com/pypa/pypi-support/issues/1422). We could point to that github repo until the issue gets sorted.

There is also a newer implementation of mustache, https://github.com/noahmorrison/chevron, but it seems less used/tested. However, if the situation above does not resolve in a timely manner, this may end up being the better solution.

psij-ci-setup
jameshcorbett: If the user is in a virtual/conda environment, though, they wouldn't need ``root``. I agree with Andre that ``--user`` could be confusing if you find that the packages actually weren't installed to your venv. And I don't think it's unreasonable to ask that anyone running the tests be in a venv.

If it's necessary we could always try a regular install and then on failure emit a warning and fall back to a user install.

hategan: > I don't think it's unreasonable to ask that anyone running the tests be in a venv.

I worry that requiring things will tend to deter people from running the tests.

Perhaps it is wiser to attempt to detect if inside a virtual/conda env and disable "--user" for those cases.

hategan: One of the difficulties I can see with venv and conda is that running things from cron may not be that straightforward, since the relevant environment may not be activated. Thoughts?

Makefile
hategan: Shouldn't we use the more general solution?

psij-ci-run
hategan: I updated this for both python and pip

psij-ci-setup
jameshcorbett: That is a good point about cron. Is there maybe a way to associate the scripts with a venv, something like "I want this script to run every TIME in venv PATH"?

hategan: So I wrote a quick update to the scripts to detect if running inside conda or venv, save the venv path or conda env name, and issue the relevant activate command before invoking the tests. I want to refine and test it a bit before committing, but it should take care of the issue.

In principle the idea is that you should be able to run the setup script straight up, under a venv, or under conda, and it will do the right thing to restore the respective environment when running the tests. We'll have to see how well it works in practice, but it should tick all the boxes.

Makefile
andre-merzky: I apologize - that comment was not meant for this thread, I somehow screwed up...

> Can I assume that all systems have Python 3 as python3?
Yes, I believe that this is the case, I did not see that not being the case anywhere (we use a fixed `#!/usr/bin/env python3` shebang on all our scripts).

andre-merzky (approved):
iteration of review - thanks.

jameshcorbett:
I figured I would move the pystache discussion to #68 since I don't think we should try to resolve it in this PR.

jameshcorbett:
Do you think you could rebase with main so we can check whether flake8 passes?

jameshcorbett:
Merging after agreement in the weekly meeting.

PR #56, opened 2021-10-22
LSF executor

jameshcorbett:
Adds an LSF executor for machines like Lassen, Sierra, and Summit. 

I gave it some basic tests on Lassen, and I'll test it on Summit soon as well.

Since `bsub` takes its batch script on `stdin`, I had to make some minor changes to existing code.

Also there is currently no support for the `cores_per_process` or `processes_per_node` options since I couldn't figure out how to implement them. There don't seem to be obvious BSUB options that correspond.


hategan:
> Since `bsub` takes its batch script on `stdin`, I had to make some minor changes to existing code.

Is this true though? https://hpc.llnl.gov/banks-jobs/running-jobs/lsf-user-manual suggests that you can also pass the script as an argument.

Even so, I can think of a few solutions that don't involve changing every existing batch executor:
- `['/bin/bash', '-c', 'bsub < %s' % submit_script]`
- writing a small wrapper script that does the above and invoking that

jameshcorbett:
Good catch, thanks. I thought I had checked that. Dropped the first commit.

hategan (approved):
Looks nice and clean.

Issue #44, opened 2021-07-16
Handling executors with dependencies


SteVwonder:
Currently, if you try to run on a system without flux or radical installed, you will get a runtime error, even if you are just attempting to use the local executor.  The radical one is a relatively easy fix since you can `pip install -r requirements-connector-radical.txt`, but no such PyPI package or requirements file exists for Flux.  Both Flux and radical (I think) are limited to Linux only, so they preclude local mac development.  As we expand the set of executors that we support, I imagine the list of overall requirements will continue to grow.

In #43 , I initially attempted to solve this issue by A) removing the explicit imports of the `flux` and `radical` executors in the `` and B) wrapping with try/except the implicit import of those executors by the plugin discovery mechanism.   For part B, if the flux or radical import failed, it was silently ignored.

@hategan rightfully pointed out that:
> I think removing the requirement for Flux to be installed and importable is fine. But I worry about the error being silently ignored. Both legitimate errors in importing flux classes as well as flux not being installed are presented to the user as the absence of the flux executor. This could potentially send users on very wasteful wild goose chases.

and then suggested:
> either moving executors with heavy dependencies to their own, separately installable packages, as I think we might have discussed as a possibility in the past. Alternatively, save the import exception and throw it when the user attempts to instantiate the flux executor. I think the latter can be made as a generic solution in the mechanism used to discover/instantiate executors.

mturilli:
Blocked by #15. Once renaming happened, @andre-merzky and @hategan should be ready to start to work on this.

SteVwonder:
In a previous Friday ☎️ call, we discussed splitting the `flux` and `radical` executors out into their own repos. This would allow the core of psi-j to be installed and run without a dependency on, say, Flux.  It would also facilitate (by providing a template to work off of) external contributors creating repos in their own organizations that provide psi-j executors.  The downside is that there would be a lot of replication of code/effort across the various executor repos in setting up CI, linting, tests, packaging, etc.

In the recent Tuesday Exaworks all-team call, @kylechard suggested an alternative: keeping everything in the same repo but still publishing the various executors as separate packages on PyPI.  This would cut down on some of the duplicate effort while still reducing the dependencies on the core of psi-j and allowing users to pick-and-choose the executors they want.

[...]

PR #61, opened 2021-10-22
Remove flux hard dependency in FluxJobExecutor

jameshcorbett:
This doesn't need to be a permanent fix, but it is pretty annoying to always have to comment out or otherwise deal with the ``flux`` ImportError. Now an ``ImportError`` is only raised on instantiation of the ``FluxJobExecutor``, which I think is a decent workaround for #44.

### DIFF
--- a/src/psij/executors/flux.py
+++ b/src/psij/executors/flux.py
@@ -26,8 +26,14 @@ class FluxJobExecutor(JobExecutor):
     A job executor that runs jobs via Flux.
     """
 
-    import radical.utils as _ru
-    import flux as _flux
+    try:
+        import radical.utils as _ru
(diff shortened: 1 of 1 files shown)


hategan:
So, awesome job on doing this and sorry for not doing it myself earlier.
That said, do you think it's maybe wise to generalize this a bit by doing the try/except for all executors in `psij.executors.__init__.py`?

jameshcorbett:
Like go around to all of them and only make the ones available that can be imported/instantiated without error? Isn't that kind of what Stephen suggested in #44? I was thinking this might be better because the executor is always 'there', and the error you get is a little more transparent.

PR #64, opened 2021-10-25
Fix mypy typing errors with Flux and LSF executors

jameshcorbett:
Fixes typing errors I introduced in #56 and #61 . (Also removes one redundant line of code).


hategan:
Thanks!

PR #70, opened 2021-11-03
Fix setuptools develop installs

### LINKED ISSUE #53: `setup.py` doesn't build scripts when using a development install
When I ran `pip install -e .` to get a development install of PSI-J, I found that the scripts in `src/psij/launchers/scripts` weren't built like they should be, and so all of my `Job`s failed. The scripts aren't built because the `build` command is not executed (`develop` is instead, I think) and so our `CustomBuildCommand` defined in `setup.py` is never run.

As a fix, we could add a `CustomDevelopCommand` and refer to it in the `cmdclass` argument to `setup()`. But my preferred fix would be to remove the `cmdclass` stuff entirely and to instead break out the script-creation logic into some free functions that run before `setup()`. I prefer that approach because I find the `setuptools/distutils` stuff a little opaque.

jameshcorbett:
Fixes #53.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,9 +2,9 @@ import distutils
 import io
 import pathlib
 
-from setuptools import setup, find_packages
-from distutils.cmd import Command
-from distutils.command.build import build
+from setuptools import setup, find_packages, Command
+from setuptools.command.develop import develop
+from setuptools.command.install import install
 
 
 class BuildLauncherScriptsCommand(Command):
@@ -41,9 +41,18 @@ class BuildLauncherScriptsCommand(Command):
                 dest.write(line)
 
 
-class CustomBuildCommand(build):
-    def run(self) -> None:
-        # build launcher scripts first
+class CustomDevelopCommand(develop):
+    """Post-installation for development mode."""
+
+    def run(self):
+        BuildLauncherScriptsCommand(self.distribution).run()
+        super().run()
+
+
+class CustomInstallCommand(install):
+    """Post-installation for installation mode."""
+
+    def run(self):
         BuildLauncherScriptsCommand(self.distribution).run()
         super().run()
 
@@ -89,6 +98,7 @@ if __name__ == '__main__':
 
         cmdclass={
             'launcher_scripts': BuildLauncherScriptsCommand,
-            'build': CustomBuildCommand,
+            'install': CustomInstallCommand,
+            'develop': CustomDevelopCommand,
         },
     )


andre-merzky (approved):
LGTM

benclifford:
Commit 2472559c51349522784fe5c25fee36c1c6f8be56 in this PR breaks my install.

I'm installing like this:

> pip install ../psi-j-python/ 

which I think should be an acceptable thing to do.

I get this error with the local executor called from my parsl provider prototype:

```
job ff93e4ad-aeb4-4b0a-b360-e09cb8314567 status JobStatus[FAILED, time=1636382664.678896, message=/bin/bash: /home/benc/parsl/exaworks/virtualenv-3.9/lib/python3.9/site-packages/psij/launchers/scripts/single_launch.sh: No such file or directory, exit_code=127]
```

I haven't really dug into what's missing/causing this yet, but I'm adding this comment in case it's obvious to someone else right away.

benclifford:
`pip install -e ../psi-j-python` works for me so this isn't blocking ongoing work for me, but it may be of more relevance when I want to have a user do an install from a network source.

hategan:
So these launcher scripts combined with python's silly install workflow may have been the single most annoying thing in this project. And, to make matters worse, the current solution is pretty bad if running on a system in which the installed package is not on a shared fs and with a stable absolute path.

So I'm thinking maybe:
1. We should move these launcher scripts to the loosely defined work dir (the one used by queuing system executors to write submit scripts and output files)
2. Build said scripts lazily. This may need some file-based locking or we can do per-process unique scripts.

Thoughts?

jameshcorbett:
Can you give more detail about the shared fs / stable absolute path issues? I don't understand. Although I agree that the installation can get pretty annoying, this particular problem was just a (somewhat subtle) bug I introduced by being careless.

hategan:
> Can you give more detail about the shared fs / stable absolute path issues?

There is no guarantee that the location where PSI/J is installed on the head node is accessible from service/compute nodes, which run the meat of the submit script, including the line that invokes the launcher. It is also possible, but unlikely, that, if that path is accessible from the CNs, it may be mounted in a different location.

The point is, if we want a robust solution that is a turn-key solution to most clusters, we may want to put those launcher scripts in a location that is more likely to be accessible to both head node and compute nodes.

andre-merzky:
I think I made the proposal in the past can't find the discussion about it, so can't remember how it was resolved.  Sorry for the repetition thus :-/  :

I would suggest to not use the preprocessor on the shell scripts.  The only purpose if to resolve the `#include <lib.sh>` statements, but the shell already has the `dot` operator to do the very same.  So `lib.sh` could be installed (preferably with a more unique name like `psij_utils.sh` in `$PYTHON_PREFIX/bin/`, and the submission scripts could begin with a `. psij_utils.sh` directive (the `dot` operator searches `$PATH` for include candidates).  

We use that mechanism in our own code without any problems.

jameshcorbett:
I like that idea @andre-merzky . (We have a bit of a parallel discussion going on in #79 so maybe we are spreading the discussion across too many issues.) I assume there is a nice way to do that with setuptools that works with develop installs?

andre-merzky:
@jameshcorbett : yeah, seems to work ok with develop installs:
<detail>

```sh
(ve_test)  $ which test.sh
(ve_test)  $ cat bin/test.sh
#!/bin/sh

export FOO=FOO
. test_lib.sh
echo "FOO=$FOO ($(which test_lib.sh))"

(ve_test)  $ cat bin/test_lib.sh

export FOO=BAR

(ve_test)  $ grep -C 1 test.sh setup.py
    'scripts'            : [
                            'bin/test.sh',
                            'bin/test_lib.sh',

(ve_test)  $ pip install -e .
[...]
  Running setup.py develop for radical.pilot
Successfully installed radical.pilot

(ve_test)  $ test.sh
FOO=BAR (/mnt/home/merzky/radical/radical.pilot.devel/ve_test/bin/test_lib.sh)

```

</detail>

andre-merzky:
> We have a bit of a parallel discussion going on in #79 so maybe we are spreading the discussion across too many issues.

Yeah, that would have fit better there, thanks for the cross-link!

PR #75, opened 2021-11-05
Use pystache fork for now and also add all relevant dependencies to r…

hategan:
…equirements-dev

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,10 +1,22 @@
+# Runtime dependencies
+filelock
+psutil
+# pystache
+https://github.com/ExaWorks/pystache/archive/refs/tags/v0.5.5-ew.tar.gz
+
+# Docs
 six
 Sphinx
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
LGTM although there are some conflicts abd I will let @andre-merzky approve as well. Also the other day I learned a neat trick for requirements files: https://stackoverflow.com/questions/11704287/split-requirements-files-in-pip so we don't have to list ``psutil`` and so on in both files.

hategan:
> although there are some conflicts

Oops. Sorry. Fixed.

hategan:
> Also the other day I learned a neat trick for requirements files

Yes, we should probably use that.

andre-merzky (approved):
LGTM

hategan:
BTW, https://github.com/PennyDreadfulMTG/pystache has now made a release (not pypi yet, just github) and added a tag. So we may want to switch to that.

PR #76, opened 2021-11-06
Some variables were inconsistent in that they sometimes did and somet…

hategan:
…imes didn't have trailing underscores


jameshcorbett (approved):
LGTM

PR #78, opened 2021-11-06
Logging times in utc

hategan:
This applies to launcher logs and logs when running tests. We may want to have the launcher logs follow whatever timezone is used for the normal logs in the future.


andre-merzky (approved):
LGTM

jameshcorbett (approved):

PR #79, opened 2021-11-08
Fix installation bug

jameshcorbett:
Fixes the issue @benclifford found [here](https://github.com/ExaWorks/psi-j-python/pull/70#issuecomment-963232614).

When pip-installing the package from a clean repository, the
launcher scripts were not moved into the install directory.
This likely has to do with the timing of the 'install' class's
'run' method (i.e. that method probably runs after all files are
expected to be built). If you install twice, the launcher scripts
will be present from the first install, and so everything will
work as expected---this is why a clean repository is necessary
to reproduce the issue.

Resolve the problem by moving the script-building into the
build_py command.

Development installs (``pip install -e``) don't have this issue because there isn't any install directory to move the scripts to.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import pathlib
 
 from setuptools import setup, find_packages, Command
 from setuptools.command.develop import develop
-from setuptools.command.install import install
+from setuptools.command.build_py import build_py
 
 
 class BuildLauncherScriptsCommand(Command):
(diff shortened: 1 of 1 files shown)


jameshcorbett:
@benclifford could you verify that this resolves your problem? Just as a sanity check.

benclifford:
This works in my environment

benclifford:
(This was slightly frustrating to test, as installing `main` with `pip -e` pollutes the source directory in such a way that `pip` without `-e` then works)

jameshcorbett:
> (This was slightly frustrating to test, as installing `main` with `pip -e` pollutes the source directory in such a way that `pip` without `-e` then works)

I agree, although if you install ``main`` with no ``-e`` it will pollute the source directory in exactly the same way, so that if you then re-install you will get a working install. That's not really due to this PR but just due to the way we construct the launcher scripts by copying ``lib.sh`` into various files. An easy fix at the cost of higher maintenance would be to get rid of ``lib.sh`` and add the launcher scripts with all their repeated code to git. Or we could do something so that the launcher scripts just source ``lib.sh``.

hategan:
> An easy fix at the cost of higher maintenance would be to get rid of `lib.sh` and add the launcher scripts with all their repeated code to git.

We could add git hooks to do this automatically and cut down on the maintenance.

jameshcorbett:
Closed in favor of #82

PR #80, opened 2021-11-10
Kill whole process tree when canceling local job

hategan:


benclifford (approved):

benclifford:
I've tested this PR against some parsl+psij connection code that I've prototyped, which was my original motivation for exploring this previously-hanging cancel behaviour. With this PR, my parsl+psij+local is able to shut down without hanging.

jameshcorbett (approved):
LGTM

PR #81, opened 2021-11-12
Fix custom attributes in mustache templates.

hategan:


andre-merzky (commented):
some short documentation and / or a small test would be great.  Or is there a ticket to reference?  I am not sure what this PR about I'm afraid...

benclifford:
The immediate motivation for this was from me testing the slurm provider at NERSC where a custom attribute specifying node type is mandatory.

With this PR I can write:

```
JobAttributes(custom_attributes=
  {"slurm":[
    {"key": "constraint", "value": "knl"},
    {"key":"foo", "value":"bar"}]
  }
)
```

resulting in these lines:

```
#SBATCH --constraint="knl"
#SBATCH --foo="bar"
```

into my slurm job submission file.

That key/value syntax is ugly but it is enough for me to move me along in my testing/development.

benclifford:
Looks like I can write this too:
```
JobAttributes(custom_attributes={"slurm.constraint":"knl", "slurm.another":"othervalue"}),
```
for a better key value experience.

hategan:
> some short documentation and / or a small test would be great. Or is there a ticket to reference? I am not sure what this PR about I'm afraid...

Sorry, my bad. Basically, what Ben says. This is a follow-up to a discussion on slack. The slurm template contains this snippet:

```    
    {{#custom_attributes.slurm}}
#SBATCH --{{key}}="{{value}}"
    {{/custom_attributes.slurm}}
```

It does not work with the way custom attributes are meant to be specified in the spec/lib. This PR makes the above snippet work correctly by setting the relevant keys in the template context dictionary and in the relevant form.

jameshcorbett:
This might be a PSI/J spec thing but I wonder if it might be better to let users specify the entire string? E.g. ``"#SBATCH --foo=bar`` rather than``{"foo": "bar"``. They're already going to need to be familiar with the batch system, and I think making users provide the entire string might be a little more transparent and simpler. I can see a lot of users mistakenly providing ``{"--foo": "bar"``. Or wanting to provide options that take a single dash rather than two: ``"#SBATCH -fbar``.

andre-merzky:
Thanks for the context all - I missed out of that slack discussion.

> but I wonder if it might be better to let users specify the entire string?

That would break the abstraction a bit: the user would need to change job description when moving from, say, Slurm to PBS.  Well, then again the key is unlikely to be valid on the PBS machine anyway - but in general I would tend to stay away from backend specifica as much as possible.  No strong opinion, just my $0.05 on this point...

andre-merzky (approved):
Code LGTM, thanks

hategan:
>> I wonder if it might be better to let users specify the entire string? E.g. `"#SBATCH --foo=bar` rather than`{"foo": "bar"`.

> That would break the abstraction a bit: the user would need to change job description when moving from, say, Slurm to PBS.

An executor implementation may choose to not use submit scripts but pass these as command line arguments to, e.g., sbatch, in which case the option would not have the "#SBATCH " part. So we may want this to come with some implicit information.

hategan:
> Or wanting to provide options that take a single dash rather than two: `"#SBATCH -fbar`

Good point. I do believe in most cases the short options are a subset of the long options, but it's still a possibility. It's also possible to have a flag, such as `--x`, without the `=value` part, which is not something that is handled right now. The spec says these are key value pairs, but otherwise it's entirely up to the executor implementation what it does with them. So maybe we can find something that could cover more scenarios. But maybe that's a separate issue.

hategan:
Well, I guess the idea of different SLURM implementations being able to recognize the same options is noble, but somewhat contradicts "it's entirely up to the executor implementation what it does with them."

jameshcorbett (approved):
LGTM and I'm happy to merge if you'd like to continue the discussion elsewhere (just let me know).

> So we may want this to come with some implicit information.

What about only making the ``#SBATCH`` implicit so that the user would have to write ``"--foo=bar"``? That would work for both sbatch command-line options and in a submit script. That's just Slurm (and LSF, I think) though so I'm not sure if it's a general improvement.

hategan:
> What about only making the `#SBATCH` implicit so that the user would have to write `"--foo=bar"`?

In contexts that I can think of at the moment (pytest/conf, gettopt) the API tends to treat the dashes as a command-line only occurrence. For example, when you want to get the value corresponding to "--foo" on the command line, you say `config.getoption("foo")`, or when you specify the options to getopt-long, you exclude the dashes. It's somewhat similar in argparse, in that you get the values without the dashes, but when you specify the options you need the full form, since argparse allows you to specify and distinguishes (maybe?) between "-foo" and "--foo". There are reasons for this, which is that the command line does not otherwise have enough structure to distinguish between positional and named arguments, a problem that does not generally exist in an API.

So maybe let's leave it as is for now.

jameshcorbett:
That works for me.

benclifford:
With my user support hat on, my main concern with any form of passing options onto an underlying system is this scenario:

A user (who knows about Science, and nothing about execution) and an HPC sysadmin (who knows about the local execution system, and nothing more) conspire to execute a program.

The HPC sysadmin sees an error and gives the solution "add `#SLURM --constraints=baz`" to your sbatch submission file.

The user (who knows nothing about execution) is unable to make sense of this meaningless instruction: they don't even know what an sbatch is, let alone whether it has a file or not.

Really the important thing for me in that use case is that it needs to be very clear how either of the intelligent-but-ignorant participants above can understand how to translate such an instruction (or other similar documentation) into the appropriate psi-j incantation.

My feeling from ops experience is that this is a documentation issue, rather than a debate-API-forms issue.

hategan:
Thanks Ben.

> My feeling from ops experience is that this is a documentation issue, rather than a debate-API-forms issue.

We'll have to carefully document the various aspects of this, and that's beyond the API docs that we are somewhat having now.

But we should also make parallel efforts to address these complexities: making things "intuitive", things like having repositories of configurations, etc.

PR #83, opened 2021-11-16
Testing updates

hategan:
This PR contains a bunch of more-or-less minor updates to testing:
- always check out a current ci-runner from main
- move non-venv non-conda dep installs from --user to .packages inside the testing setup dir
- install/upgrade dependencies on every branch that is tested
- add a requirements-tests.txt, which installs only dependencies needed for testing

### DIFF
--- a/psij-ci-setup
+++ b/psij-ci-setup
@@ -1,20 +1,21 @@
 #!/bin/bash
 
+set -e
+
 if pip --version 2>&1 | egrep -q 'python 3\..*' >/dev/null 2>&1 ; then
     PIP="pip"
 else
     PIP="pip3"
 fi
 
+MODE="plain"
 
 if [ "$VIRTUAL_ENV" != "" ]; then
-    VENV=1
-    NOUSER=1
+    MODE="venv"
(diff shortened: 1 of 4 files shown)


psij-ci-setup
```diff
@@ -51,11 +58,19 @@ done
...
+if [ "$MODE" != "plain" ]; then
+    OUT=`$PIP install -r requirements-tests.txt 2>&1`
 else
-    $PIP install --user -r requirements-dev.txt
+    OUT=`$PIP install --target .packages --upgrade -r requirements-tests.txt 2>&1`
+fi
+
+if [ "$?" != "0" ]; then
+    echo "FAILED"
+    echo $OUT
```
andre-merzky, line 71: Should we exit here?

andre-merzky (approved):
One minor comment, otherwise LGTM - thanks!

psij-ci-setup
hategan: We should.

jameshcorbett (approved):
LGTM

PR #84, opened 2021-12-01
Invoke launcher with actual deployed path rather than the source path.

hategan:
On non-clusters, the existing code works. On clusters, when tests are
run from a non-shared FS, the path inside the source to the launchers
is not visible from CNs, so things fail.


hategan:
I've no idea what happened here.

PR #85, opened 2021-12-01
Invoke launcher with actual deployed path rather than the source path.

hategan:
On non-clusters, the existing code works. On clusters, when tests are
run from a non-shared FS, the path inside the source to the launchers
is not visible from CNs, so things fail.

### DIFF
--- a/src/psij/launchers/script_based_launcher.py
+++ b/src/psij/launchers/script_based_launcher.py
@@ -98,14 +98,14 @@ class ScriptBasedLauncher(Launcher):
                 return
 
             self._deploy_file(Path(__file__).parent / 'scripts' / 'launcher_lib.sh')
-            self._deploy_file(self._script_path)
+            self._deployed_script_path = self._deploy_file(self._script_path)
 
-    def _deploy_file(self, path: Path) -> None:
+    def _deploy_file(self, path: Path) -> Path:
         dst_dir = self.config.work_directory
         dst_dir.mkdir(parents=True, exist_ok=True)
         dst_path = dst_dir / path.name
         if dst_path.exists():
-            return
+            return dst_path
         tmp_prefix = secrets.token_hex() + '_'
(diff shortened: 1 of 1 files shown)


src/psij/launchers/script_based_launcher.py
```diff
@@ -98,26 +98,29 @@ def _ensure_launcher_deployed(self) -> None:
...
         tmp_path = dst_dir / (tmp_prefix + path.name)
         shutil.copy(path, tmp_path)
         try:
             # this appears to use os.rename, although I'd wish pathlib docs would state this
             # explicitly, since the docs for Path.rename mention nothing of the exceptions thrown
             tmp_path.rename(dst_path)
+            return dst_path
         except FileExistsError:
             # thrown in Windows if the path already exists, which is fine if the destination is a
             # file; we were recing another process
```
andre-merzky, line 119: recing -> racing?

andre-merzky (approved):
LGTM.  I don't understand the comment about the shared FS: if the deployment happens on the non-shared FS, then the CN will not see the files anyway, despite the full path?

src/psij/launchers/script_based_launcher.py
hategan: pfft "racing". "recing" is all the rage now.

hategan:
> if the deployment happens on the non-shared FS, then the CN will not see the files anyway, despite the full path?

Right. The test runner makes a git clone in /tmp. The current main code points to the launcher file inside that clone, which can't be seen by the actual job. See the log for the slurm:single test in https://testing.exaworks.org/run.html?site_id=midway2-login2.rcc.local&run_id=2021113020-6dcd9623096864922cabfaa0765345beb326e1866b8394db2e1cb6099cdf4b57#branch-main

andre-merzky:
Thanks for the clarification, makes sense.

PR #86, opened 2021-12-01
Added psij-ci-load, a file that can be used to set up the environment…

hategan:
… before running tests.

### DIFF
--- a/README-testing.md
+++ b/README-testing.md
@@ -19,6 +19,11 @@ running the integration tests and through `cron` (or a similar tool).
 Setting up a Cron testing job
 =============================
 
+Note: If you need to set up an environment module (such as 
+`module load python/cpython-x.y.z`) or something similar, please add the 
+relevant commands to `psij-ci-load`.
+
+
 This is the preferred way of running the tests since it allows the PSI/J
 team to keep a constant eye on the state of the library on various
 resources. To set up the Cron job, you can either use the provided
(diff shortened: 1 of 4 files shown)


andre-merzky (commented):
Syntactically and semantically this looks correct to me.  I am not sure about the use case.  It's usually in the hands ob the user to set up environment for any application or tools she is using.  I guess the rationale is to make this somehow simpler and persistent?  But unless this is a standard way to do things, the user now has *additional* work to do (changing this file for `psij`, a wrapper script for some other code, some explicit export for a third?).  

So my question is what prompted this addition?

hategan:
> So my question is what prompted this addition?

A very good question. And perhaps you know of a better solution. My familiarity with cluster environments is a bit rusty.

I was trying to set up tests on a UChicago cluster called Midway. It didn't, by default, have Python in the environment. You have to do "module load python...". Adding it as a default using "module initadd" does not appear to work. Adding it to ~/.modules also didn't. I couldn't find another way of reliably having it persistently.

If only the setup mattered, it would be fine, since it's a one time operation. However, python needs to also be in the path whenever the test executable is invoked by cron.

The user doesn't normally need to do anything about it. It's a mechanism that does not have to be used if there is a better solution, so I don't agree that the user has additional work to do. This is something to be used when there is no other solution.

benclifford:
I've used this at nersc to get their weird conda-that-isn't-quite-conda working (rather than using the psij-ci conda detection).

PR #87, opened 2021-12-01
Adds a srun launcher for Slurm.

hategan:


benclifford:
I tried this with my parsl/psij provider, replacing the use of the parsl-side srun launcher with the psij-side srun launcher.
Then I checked that a small parsl workflow on three slurm nodes still uses all three nodes.

benclifford (approved):

benclifford:
In a few situations, I've wanted to add parameters to `srun`. In parsl, that's provided as an `overrides` parameter which pastes arbitrary extra stuff into the command line. I'm not sure if this new laucher implementation would let me easily configure those options.

An example from an LSST parsl configuration: `launcher=SrunLauncher(overrides='-K0 -k --slurmd-debug=verbose'),`

PR #89, opened 2021-12-02
When trying to ssh to localhost, ensure that batch mode is requested,…

hategan:
… so that we don't get prompts, and ensure that keys are added automatically, since we implicitly trust localhost.


andre-merzky (approved):
LGTM

PR #97, opened 2021-12-07
Add an executor-dependent timeout to tests to reduce the possibility …

hategan:
…of a deadlock in an executor blocking all tests.


andre-merzky (commented):
LGTM

andre-merzky (approved):

jameshcorbett (approved):
LGTM as well. It would be great to be able to timeout every executor but alas :(. Should I go ahead and merge? (Not sure whether we are requiring the CI to pass.)

PR #99, opened 2021-12-07
OpenMPI mpirun does not like to run as root. Convince it to do so for…

hategan:
… the github ci container.


andre-merzky (approved):
LGTM

PR #100, opened 2021-12-08
Switch back to official pystache

hategan:
The PyPI pystache packages has been taken over by another developer who is now actively maintaining it. This version is also now compatible with newer versions of Python (3.6+).

Additionally, this PR reorganizes the requirements files a bit to make them consistent and reduce duplication.


andre-merzky (approved):
LGTM

PR #104, opened 2022-01-07
Dynamic executor import errors

hategan:
This is mainly a revamp of the executor/launcher loading code. The main stuff (with description) is in ad9e42451f23e6d6a09070683a90011a9a679208, and I'll reproduce the text below.

There are some tests in fd56015ca0ec8c64188a8b67404468e6b9e660bd

Plugins are now loaded indirectly, using a descriptor class. This was
needed because one could not find details about, say, an executor (such
as name and version) if the executor file failed to import. However, a
minimal descriptor is virtually guaranteed to load, providing information
about the implementation that does not depend on that implementation being
able to load/import.

Also updated the way loading works. Previously, problems could occur when
implementations existed in multiple locations, because package traversal
in python is weird: despite correctly iterating over packages from
multiple locations, python only really references the first location
loaded and only allows imports from that location. The current
implementation works according to the following rules:
- descriptors must be in the 'psij.executors' package, and the package can
be spread across multiple locations
- implementations must exist in the same location as the descriptor; for
example, a descriptor in the system dist-packages can only be used to load
an implementation from the system dist-packages, not from a user directory.
- if implementations exist in both the same location and another location
as a descriptor, only the one in the same location will be used
- here, "location" means one directory in PYTHONPATH and/or sys.path.

And last, you can now run "python -m psij plugins" to get a detailed list
of what is loaded and from where.


hategan:
A bit of a note on tests: 3 of the new tests require some directories to be added to PYTHONPATH before invoking pytest. This is done by ci_runner.py, which is always checked out from the main branch. So those tests will fail on this branch/PR until it's merged.

src/psij/_plugins.py
```diff
@@ -0,0 +1,123 @@
...
+        if spec is None:
+            raise Exception('Failed to load %s' % mod_path)
+        mod = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(mod)  # type: ignore
+        cls = getattr(mod, cls_name)
+    except Exception as ex:
+        s = str(ex)
+        if 'does_not_exist' not in s:
+            # there is a class used for testing that would print a warning message every time
+            # psij was imported; intercept that message and avoid logging it
```
benclifford, line 61: it isn't immediately clear to me what this comment means wrt the surrounding code: you're logging all exceptions from module loading except ones containing the string "does_not_exist" which you are hoping only comes from a test class and not from anything user relevant?

benclifford:
> A bit of a note on tests: 3 of the new tests require some directories to be added to PYTHONPATH before invoking pytest. This is done by ci_runner.py, which is always checked out from the main branch. So those tests will fail on this branch/PR until it's merged.

this breaks the pytest command line in README-testing.md - maybe remove that entirely as `make test` is probably enough documentation

src/psij/_plugins.py
benclifford: could remove this from the core code, and only import the deliberately-broken test case descriptor during test runs? if descriptors can be spread across multiple locations, and there's a specific extra path set already for testing?

src/psij/__init__.py
```diff
@@ -41,26 +42,49 @@ def __init__(self, name: str, registration_method: Callable[[Type[T]], None]):
...
+                im = importlib.util.module_from_spec(spec)
+                spec.loader.exec_module(im)  # type: ignore
+                full_mod_path = spec.origin
+
+                var_name = '__PSI_J_{}__'.format(_type.name.upper())
+                if hasattr(im, var_name):
+                    classes = getattr(im, var_name)
+                    logger.debug('Found module "{}" with classes {}'.format(mod.name, classes))
+                    for cls in classes:
+                        if isinstance(cls, type) and issubclass(cls, JobExecutor):
```
benclifford, line 63: everything in the repo is ported to the new descriptor format so this code path and the launcher one below doesn't fire for anything in-repo at the moment, right?

hategan: ... unless you have an old version installed on the system, which I do, and which I'm keeping for now as a weird-environment-test-case.

hategan:
> I've encountered this behaviour which is somewhere between awkardness and failure [...] path order [...]

I noticed this as I was adding the paths to ci_runner.

> 
> Where this is more a problem than an awkwardness is when I have installed `psij` into a `conda` environment, but want to add in more stuff using PYTHONPATH - for example, I installed psij in conda and tried out my pbspro executor as a separate module.

We cannot enforce/guarantee a particular order when this is used, so this is a problem.

> [...]
> 
> This doesn't feel right to me.

I agree.

> 
> The closest I know of in the Python documentation is namespace packages, here: https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages
> 
> which talks about doing things like omitting the top level **init**.py so as to make that package a "namespace package".
> 
> I don't have any practical experience with doing that, though.

Since I discovered the issue, I've been experimenting with that. It doesn't work, at least not the way things are structured now. The problem with namespace packages is that we do need `psij` to be a normal package with things done when it is loaded and that prevents us from using it as a namespace package.

What I will experiment with next is having a separate namespace package for descriptors and "mandating" separate packages for the executors themselves (e.g. `psij-executor-xyz.xyz_executor.py`.

> 
> I'm also unclear about what the package modularisation goals for the code that this PR deals with - eg what should end users be able to do (or not do) wrt separate installs of pieces?

This does try to address the separate install of pieces. However, the core loader has to be able to find the pieces, so they have to be in a more-or-less pre-determined "location".

hategan:
> What I will experiment with next is having a separate namespace package for descriptors and "mandating" separate packages for the executors themselves (e.g. `psij-executor-xyz.xyz_executor.py`.

b419aeedc4ebcb136df6570f1c24bba3492f6e80 adds the separate namespace package (`psij-descriptors`). It also contains a more detailed explanation of what and why. It appears to work in all situations I've tested.

There's also an earmark (d9028490337aa3180ef0e650957cdfe57fdf081a) which adds a branch-specific wrapper between ci_runner and pytest, so that some customization can be done on specific branches (since ci_runner is always used from the main branch).

benclifford:
ok, with separate `psij-descriptors`, my psij-in-conda, pbspro-in-PYTHONPATH test works, at least as far as getting jobs in the queue (looks like that system isn't actually running jobs at the moment...)

The directory tree so far looks like (though probably doesn't need the subdirectory structure so much now it's in its own top-level package)

```
./psijpbs/__init__.py
./psijpbs/executors/__init__.py
./psijpbs/executors/pbspro.py
./psijpbs/executors/batch/pbspro/pbspro.mustache
./psij-descriptors/pbspro_descriptor.py
```

hategan:
@benclifford good to hear.

I think it's fine to leave the exact location of the executor implementation up to whoever is writing it.

src/psij/_plugins.py
hategan: Fixed by efc7231ddae9375cdb04dd3a1a5a9d2fbcb1a2e2

src/psij/__init__.py
```diff
@@ -31,44 +31,66 @@
...
 
-def _find_plugins(path: str, type: _PluginType, ix: int) -> None:
-    if ix == len(type.package):
-        _load_plugins(path, type)
-    else:
-        for mod in pkgutil.iter_modules(path=[path]):
-            if mod.ispkg and mod.name == type.package[ix]:
-                _find_plugins(path + os.path.sep + mod.name, type, ix + 1)
+def _load_plugins(root: str, full_path: str, mod: pkgutil.ModuleInfo) -> None:
+    if not mod.ispkg and not mod.name[0] == '_':
```
benclifford, line 47: what's excluded by this `_prefix` clause?

src/psij/_descriptor.py
```diff
@@ -0,0 +1,48 @@
...
+
+class _Descriptor(object):
+    def __init__(self, name: str, version: StrictVersion, cls: str, path: Optional[str] = None):
+        self.name = name
+        self.version = version
+        self.cls = cls
+        self.path = path
+
+    def __repr__(self) -> str:
+        return '_Descriptor[name=%s, path=%s]' % (self.name, self.path)
```
benclifford, line 48: i think the __repr__ docs encourage ( rather than [ here.

src/psij/_plugins.py
```diff
@@ -0,0 +1,120 @@
...
+        s = str(ex)
+        logger.info(s)
+        exc = ex
+
+    if desc.name not in store:
+        store[desc.name] = []
+    existing = store[desc.name]
+    entry = _VersionEntry(desc.version, desc_path=desc.path, plugin_path=mod_path,
+                          ecls=cls, exc=exc)  # type: _VersionEntry[T]
+    # check if an object with this version already exists
```
benclifford, line 67: so does this code allowed registration of multiple versions of the same-named executor?

benclifford (approved):
ok I think this is ok to merge. I made some passing style comments / queries on docs but I don't think they're essential.

src/psij/__init__.py
hategan: It's a remnant from when executors were there and you maybe wanted to have a non-executor in the same package. I don't think it's needed now given that we have a special package for only descriptors.

src/psij/_descriptor.py
```diff
@@ -0,0 +1,48 @@
...
+    def __cmp__(self, other: '_VersionEntry[T]') -> int:
+        if self.version == other.version:
+            return 0
+        elif self.version < other.version:
+            return -1
+        else:
+            return 1
+
+    def __eq__(self, other: object) -> bool:
+        assert isinstance(other, _VersionEntry)
```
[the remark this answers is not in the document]
hategan: Here, yes. Should be changed. But you say "these"?

src/psij/_descriptor.py
hategan: Thank you. Will update.

src/psij/_plugins.py
hategan: Yes.

src/psij/_descriptor.py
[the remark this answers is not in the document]
hategan: Nevermind. I see.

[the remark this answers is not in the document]
hategan: Well, actually, hold on. Maybe for equality that makes sense, but for `__lt__` and `__gt__`, shouldn't it be reasonable that not everything can be compared with descriptors?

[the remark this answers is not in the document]
benclifford: I think it's this I'm thinking of:

> A rich comparison method may return the singleton NotImplemented if it does not implement the operation for a given pair of arguments.

https://docs.python.org/3/reference/datamodel.html

In practice this is maybe only going to be a problem if they're combined in interesting data structures with other types, so right now that's probably not a problem for this code.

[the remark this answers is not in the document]
hategan: Thanks. That make sense. Except maybe "the truth of (x<y or x==y) does not imply x<=y" :)

hategan:
@benclifford if you are OK with the various fixes and, seeing that there's likely no other reviewers, can you click the merge button?

jameshcorbett:
I'll take a look today and merge if things check out, if you don't mind.

hategan:
> I'll take a look today and merge if things check out, if you don't mind.

Not at all. Thank you!

jameshcorbett (approved):
LGTM, sorry for the delay. Thanks for putting this together.

PR #122, opened 2022-01-20
This logs sys.path, but removes explicit references to home directories.

hategan:


benclifford:
what is this for? logging to the central log server? or using the paths for some other purpose?

hategan:
> what is this for? logging to the central log server? or using the paths for some other purpose?

Sorry. Yes, logs to the central server the python path that the tests are running under.

benclifford:
Merged this, with the disclosure that I kinda have a bad feeling about doing clever substitutions that make it look like a setting is set to something other than what it is, from debugging frustrations in the past.

hategan:
> Merged this, with the disclosure that I kinda have a bad feeling about doing clever substitutions that make it look like a setting is set to something other than what it is, from debugging frustrations in the past.

This may have a simple solution: make it obvious. E.g., "<home>/" instead of "~/"

PR #128, opened 2022-01-27
On some OSX installations, bash has v3.2.x. That version does not rec…

hategan:
…ognize the timestamp format for printf.

So this PR changes that part to use `date` with (hopefully) a widely accepted format string.


andre-merzky (approved):
LGTM

wilke (approved):

PR #129, opened 2022-01-30
Export type annotations (PEP 561)

benclifford:
This allows user packages (in my case, the parsl provider) to be checked against type annotations of psij code, when mypy is used when developing those user packages.

This is in contrast to the existing use of mypy and type annotations in psij, which checks that psij is internally self-consistent.


hategan (approved):
They should invent the equivalent of a generic stamp of approval.

PR #132, opened 2022-02-03
Reorganized sphinx documentation a bit and added placeholders for gui…

hategan:
…des.

The basic idea is that the main doc page is now a set of links to a number of guides (installation, testing, programming, contributing).  The api docs were moved to the programming guide. The main page of the programming guide is written by hand now rather than it being just a set of ToCs.


benclifford (approved):
I didn't proof read this in depth but this is fine for getting merged for further work

PR #140, opened 2022-02-09
Add PBSPro executor

benclifford:
This PR adds a basic PBS Professional executor.

Most notably missing, this code does not pay attention to all jobspec (and descendant) fields. However, other executors also do not do that (eg slurm vs `environment`) so I don't think that should block this being merged.


src/psij/executors/batch/pbspro.py
```diff
@@ -0,0 +1,118 @@
...
+                              # very end of being ACTIVE or the very
+                              # beginning of being COMPLETED. This mapping
+                              # treats it as ACTIVE because not all final
+                              # job information has appeared in JSON yet.
+}
+
+
+class PBSProExecutorConfig(BatchSchedulerExecutorConfig):
+    """A configuration class for the PBS executor.
+    This doesn't have any fields in addition to BatchSchedulerExecutorCOnfig,
```
jameshcorbett, line 31:
```suggestion
    This doesn't have any fields in addition to BatchSchedulerExecutorConfig,
```

src/psij/executors/batch/pbspro.py
```diff
@@ -0,0 +1,118 @@
...
+            config = PBSProExecutorConfig()
+        super().__init__(url=url, config=config)
+        self.generator = TemplatedScriptGenerator(config, Path(__file__).parent / 'pbspro'
+                                                  / 'pbspro.mustache')
+
+    # Submit methods
+
+    def generate_submit_script(self, job: Job, context: Dict[str, object],
+                               submit_file: TextIO) -> None:
+        """See :proc:`~BatchSchedulerExecutor.generate_submit_script`."""
```
jameshcorbett, line 53: This may be correct but I've never seen ``:proc:`` before... Is that a real directive?

jameshcorbett (commented):
Code looks good. My preference would be for you to clean up the commits and reduce them to 1-3, but I'm not sure how the other developers feel about that. (Maybe you already intended to.)

It would be nice if you provided URLS to some reference web pages in some of the docstrings, so we know where to look if we need to make changes or add new features.

benclifford:
@jameshcorbett  > Code looks good. My preference would be for you to clean up the commits and reduce them to 1-3, but I'm not sure how the other developers feel about that. (Maybe you already intended to.)

Mostly I commit stuff as if the whole PR is going to be squash merged - so expect it to reduce to 1 if I remember to hit the right merge button...

src/psij/executors/batch/pbspro.py
benclifford: Maybe, maybe not - but it isn't used in the rest of the code base so it would be more consistent to make it :meth:

benclifford: I've made that change

benclifford:
Merging this after informal approvals elsewhere

PR #142, opened 2022-02-09
Increase mypy lower bound as 0.790 breaks

benclifford:
In other projects I've found mypy behaviour changing fast enough
that I've ended up pinning a version explicitly.


hategan (approved):
There should be a relatively low bar for bumping up versions of packages only required for dev, unless that breaks stuff.

PR #155, opened 2022-03-01
Flux executor rewrite

### LINKED ISSUE #18: Flux executor should work with existing Flux instance

### LINKED ISSUE #31: CI: add testing for Flux connector
Blocked by: https://github.com/ExaWorks/SDK/issues/47 and https://github.com/ExaWorks/psi-j-python/pull/14

### LINKED ISSUE #50: Flux executor: re-add support for `attach`
https://github.com/ExaWorks/psi-j-python/pull/42/commits/73bc66bad3668659e2950ee9761ba8236a1c6933 temporarily dropped support for `attach` in the Flux executor due to errors detected by mypy and the need for the FluxExecutor class in Flux (not to be confused with jpsi's Flux executor) to support attaching to normal Flux jobs submitted outside of the FluxExecutor interface.  Support was added with https://github.com/flux-framework/flux-core/pull/3790 but no tag has occurred since then.  Once a tagged release occurs, maybe then would be a good time to revisit this.

### LINKED ISSUE #52: Flux executor broken
I am not sure if this is a problem with my local setup, or if the state of the executor deteriorated for some other reason, but I observe this behavior on `main`:
```sh

$ ./tests/run_job.py local
Job submitted
Job done: JobStatus[COMPLETED, time=1629979965.2759883, exit_code=0]

$ ./tests/run_job.py flux
Job submitted
Job done: JobStatus[FAILED, time=1629979974.6525922, meta={'type': 'alloc', 'severity': 0, 'userid': -1, 'note': 'alloc denied due to type="unsatisfiable"', 'exit_code': None, 'message': 'alloc denied due to type="unsatisfiable"'}]
```

I did not yet dig into it, just recording the ticket for now.

PS.: the ticket referred to running the test w/i a flux instance - but that mode is not supported in main yet, there a flux instance is started internally - I edited to reflect that.

jameshcorbett:
Closes #52 
Closes #18 
Closes #31 
Closes #50

### DIFF
--- a/src/psij/executors/flux.py
+++ b/src/psij/executors/flux.py
@@ -1,45 +1,52 @@
-"""This module contains the Flux :class:`~psij.JobExecutor`."""
+"""This module contains the Flux :class:`~psij.JobExecutor`.
 
-from __future__ import annotations
+Implementation references:
+github.com/flux-framework/flux-core/blob/master/src/bindings/python/flux/job/executor.py
+flux-framework.readthedocs.io/projects/flux-core/en/latest/python/job_submission.html#the-fluxexecutor-interface
+
+Events and state transitions:
+github.com/flux-framework/rfc/blob/master/spec_21.rst
+"""
 
 import time
-import logging
 import copy
-import json
-import threading
+import warnings
+import concurrent.futures
 from functools import partial
-
 from typing import Any, Optional, Dict, List
 
-from psij import InvalidJobException
-from psij import Job, JobExecutorConfig, JobState, JobStatus, JobAttributes, JobSpec
-from psij import JobExecutor, ResourceSpecV1
+from psij import (
+    Job,
+    JobExecutorConfig,
+    JobState,
+    JobStatus,
+    JobExecutor,
+    InvalidJobException,
+    ResourceSpec,
+    ResourceSpecV1,
+)
 
-logger = logging.getLogger(__name__)
-# logging.basicConfig(level='DEBUG', force=True, filename='flux_connector.log')
+import flux
+import flux.job
 
 
 class FluxJobExecutor(JobExecutor):
     """A job executor that runs jobs via Flux."""
 
-    # map flux states to jpsi states.
-    _state_map = {'NEW': JobState.QUEUED,
-                  'DEPEND': None,
-                  'SCHED': None,
-                  'RUN': JobState.ACTIVE,
-                  'CLEANUP': None,
-                  'INACTIVE': JobState.COMPLETED}
-    _event_map = {'submit': JobState.QUEUED,
-                  # 'alloc': None,
-                  'start': JobState.ACTIVE,
-                  'finish': JobState.COMPLETED,
-                  # 'release': None,
-                  # 'free': None,
-                  # 'clean': None,
-                  'exception': JobState.FAILED}
+    _event_map = {
+        # 'submit': JobState.QUEUED,
+        # 'alloc': None,
+        "start": JobState.ACTIVE,
+        # 'finish': JobState.COMPLETED,
+        # 'release': None,
+        # 'free': None,
+        # 'clean': None,
+        # 'exception': JobState.FAILED
+    }
 
-    def __init__(self, url: Optional[str] = None,
-                 config: Optional[JobExecutorConfig] = None) -> None:
+    def __init__(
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -118,6 +118,8 @@ def _translate_executor(config: Dict[str, str], executor: str) -> List[str]:
     if executor == 'auto_q':
         if config.option.environment.get('has_slurm'):
             return ['slurm']
+        elif config.option.environment.get('has_flux'):
+            return 'flux'
```
andre-merzky, line 122: This likely should return `['flux']`?

jameshcorbett: Duh, thanks!

src/psij/executors/flux.py
```diff
@@ -50,175 +57,110 @@ def __init__(self, url: Optional[str] = None,
...
+            warnings.warn(
+                "Flux does not support exclusive_node_use=True, ignoring...",
+                UserWarning,
+            )
+        if resources.processes_per_node:
+            raise InvalidJobException("Flux does not support processes_per_node")
+        if not job.spec.executable:
+            raise InvalidJobException("Job must have an executable")
+        argv = list(job.spec.arguments) if job.spec.arguments else []
+        argv.insert(0, job.spec.executable)
```
hategan, line 141: `[job.spec.executable] + job.spec.arguments` (or similar) will save you one O(n) operation.

src/psij/executors/flux.py
```diff
@@ -50,175 +57,110 @@ def __init__(self, url: Optional[str] = None,
...
+        )
+        if job.spec.stdout_path:
+            flux_jobspec.stdout = job.spec.stdout_path
+        if job.spec.stdin_path:
+            flux_jobspec.stdin = job.spec.stdin_path
+        if job.spec.stderr_path:
+            flux.jobspec.stderr = job.spec.stderr_path
+        flux_jobspec.duration = job.spec.attributes.duration.total_seconds()
+        fut = self._flux_executor.submit(flux_jobspec)
+        self._add_flux_callbacks(job, fut)
```
hategan, line 157: That looks like a bit of a race condition.

hategan (approved):
I had some minor comments, except the exit_code check in the tests, which I think is important.

src/psij/executors/flux.py
jameshcorbett: That's true, will fix, my bad assumption about Python list performance. Verified with python wiki.

hategan: Won't make much of a difference, since you already have an O(n) and n is small, so feel free to keep, but I thought the alternative would also look nicer.

tests/test_executor.py
```diff
@@ -98,8 +98,8 @@ def test_missing_executable(execparams: ExecutorTestParams) -> None:
         status = job.wait(timeout=_get_timeout(execparams))
         assert status is not None
         assert status.state == JobState.FAILED
-        assert status.exit_code is not None
-        assert status.exit_code != 0
+        if status.exit_code is not None:
```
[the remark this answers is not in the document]
jameshcorbett: That's a different test, this one tests what happens when you run '/bin/no_such_file_or_directory'. In that case Flux throws an exception rather than returning the exit code. It is possible (with a little more work) to fetch the exit code (probably -127) from the job, but I thought it was reasonable not to. Let me know what you think though and I can add it.

src/psij/executors/flux.py
jameshcorbett: It isn't. The executor stores all the events it has seen and if you add a callback after the executor has handled it, the callback will be invoked immediately. Otherwise we (and concurrent.futures which operates the same way) would be screwed, because you can't get a handle to a future until you've submitted a job.

tests/test_executor.py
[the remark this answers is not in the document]
hategan: My bad. Yeah, I think that's reasonable. Ignore my initial comment.

src/psij/executors/flux.py
hategan: - [x] &nbsp;

src/psij/executors/flux.py
jameshcorbett: Yeah, I am a bit to lazy to rebase right now, so I will just merge. If I remember I will make the change later.

PR #156, opened 2022-03-03
Added flux to mocked imports when building docs.

hategan:


hategan:
> LGTM but can you give a brief explanation? Without this does the docs build fail? Or can it just not import/document the Flux executor?

The docs build fails since sphinx tries to directly import `psij.executors.flux`.

PR #157, opened 2022-03-03
Add flake8 docstrings

hategan:
This enables additional checks for docstring formatting.

The PR also contains some docstring fixes to make it actually work.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,6 +9,7 @@
 mypy >=0.931
 pytest >= 6.2.0
 flake8
+flake8-docstrings
 autopep8
 types-requests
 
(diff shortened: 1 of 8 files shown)


jameshcorbett (approved):
Looks great, especially the new ``Descriptor`` docstring. Does having flake8-docstring installed make docstring checks run automatically with flake8?

hategan:
> Does having flake8-docstring installed make docstring checks run automatically with flake8?

I'm going to assume so. I'm doing this because I noticed that I was getting errors locally when the git tests would pass. So I set to investigate why, and I noticed that flake8 had that plugin installed. I suppose we can test it.

hategan:
Hmm, not sure why only two tests are running.

hategan:
>> Does having flake8-docstring installed make docstring checks run automatically with flake8?
> I suppose we can test it.

Looks like it.

PR #158, opened 2022-03-03
Moved batch-test executor to tests/ directory. 

### LINKED ISSUE #130: Move batch-test executor to testing directory.
The batch-test executor is used to test the abstract classes for batch schedulers. It is only used by tests, hence, like the always-loads and never-loads executors, should belong in the tests directory.

hategan:
Fixes #130.


andre-merzky (approved):
LGTM

jameshcorbett (approved):
LGTM as well.

PR #159, opened 2022-03-04
Moved `launcher.py` into the main `psij` directory. 

### LINKED ISSUE #131: Inconsistent location between JobExecutor and Launcher
It seems to me that JobExecutor and Launcher are similar concepts. Yet, they live in different packages  (`psij` vs. `psij.launchers`) despite everything else seemingly being the same (e.g., there is a `psij.executors` package, both Launcher and JobExecutor are first class citizens in the spec, etc.).

We should either move `JobExecutor` into `psij.executors` or `Launcher` into `psij`.

hategan:
Fixes #131.

### DIFF
--- a/docs/programming.rst
+++ b/docs/programming.rst
@@ -29,7 +29,7 @@ The core API contains the central components of PSI/J. The core API classes are 
 :class:`~psij.job_executor.JobExecutor`
     The base class for a concrete method of executing a job, such as Slurm, or PBS.
 
-:class:`~psij.launchers.launcher.Launcher`
+:class:`~psij.launcher.Launcher`
     A base class representing a method of actually launching a job once resources are allocated for it, such as `mpirun`, `srun`, etc.
 
 :py:mod:`~psij.exceptions`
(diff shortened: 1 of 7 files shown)


andre-merzky (approved):
LGTM, thanks.

jameshcorbett (commented):
I don't like having two modules/packages with such similar names at the same level: ``psij.launchers`` and ``psij.launcher``. I just think it gets confusing. Couldn't we keep the launcher class in ``psij.launchers.launcher.Launcher`` but export it as ``psij.Launcher``?

hategan:
> I don't like having two modules/packages with such similar names at the same level: `psij.launchers` and `psij.launcher`. I just think it gets confusing. Couldn't we keep the launcher class in `psij.launchers.launcher.Launcher` but export it as `psij.Launcher`?

Would it make sense to rename launcher.py and Launcher to job_launcher.py and JobLauncher, respectively? I mean the point here was to have some symmetry with executors.

jameshcorbett:
> Would it make sense to rename launcher.py and Launcher to job_launcher.py and JobLauncher, respectively? I mean the point here was to have some symmetry with executors.

I think just renaming the module is fine, renaming the class is overkill.

docs/programming.rst
```diff
@@ -29,7 +29,7 @@ The core API contains the central components of PSI/J. The core API classes are
 :class:`~psij.job_executor.JobExecutor`
     The base class for a concrete method of executing a job, such as Slurm, or PBS.
 
-:class:`~psij.launchers.launcher.Launcher`
+:class:`~psij.launcher.Launcher`
```
jameshcorbett, line 32: Looks like this file needs updating.

jameshcorbett (approved):
LGTM, one minor comment below.

docs/programming.rst
hategan: Thanks for spotting it. I somehow missed it when adding files to the commit.

PR #162, opened 2022-03-07
Wilke/command line tool

### LINKED ISSUE #103: Provide a command line tool to run jobs
Swift/T, and other non-python or shell based projects, in order to use PSI/J, would need a command-line interface to running jobs. This issue is here to discuss the requirements, come up with a design, and track the implementation.

- an initial task would be to attempt to gather requirements from teams that are potential users of such a feature (@j-woz mentioned one such team).
- an initial rough design is based on the idea that one needs to communicate a JobSpec (and properties) and executor + config from one process to the other; this would be done through some serialization mechanism, which could take the form of:
    - command line arguments (e.g., `--spec.attributes.duration=...`)
    - json
- we talked about a sync vs async interface and concluded that the biggest benefit/cost ratio would be achieved with a sync implementation, since more complex projects would use higher level tooling; for reference:
    - sync here means `run-job ...` which exits when the job completes
    - async implies a set of tools, such as `submit-job`, `cancel-job`, `job-status`; there is considerably more complexity in implementing these, at least in layer 0 (basically these would roughly be a local version of layer 1)

wilke:
closes #103


src/psij/job_spec.py
```diff
@@ -132,7 +132,10 @@ def to_dict(self) -> Dict[str, Any]:
             }
             for k, v in self.attributes.__dict__.items():
                 if k in ['duration', 'queue_name', 'project_name', 'reservation_id']:
-                    d['attributes'][k] = str(v)
+                    if v:
+                        d['attributes'][k] = str(v)
```
jameshcorbett, line 136: What's the reason for this change?

scripts/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This directory contains examples, helper scripts and the console for PSIJ.
+
+## psij-console
+
+The console takes an exported JobSpec document and either validates or executes it.
+
+
+```
+usage: psij-consol [-h] [-v] [--debug] {validate,run} ...
```
jameshcorbett, line 11: I'm not a huge fan of this name but I can't think of anything better, and we can always change it later I guess.

scripts/psij-consol.py
```diff
@@ -0,0 +1,95 @@
...
+validate_parser.add_argument("file", help="JobSpec file")
+execute_parser = subparser.add_parser("run", help='execute JobSpec file')
+execute_parser.add_argument("file", help="JobSpec file")
+execute_parser.add_argument("-j", 
+                            "--job-executor",
+                            dest = "executor",
+                            choices = ["cobalt"
+                                       , "local"
+                                       , "batch-test"
+                                       , "flux"
```
jameshcorbett, line 26: The comma should be after the entry, and not before the next entry? But if you like it this way, that's OK I guess. (I really don't like it though.)

scripts/psij-consol.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+
+parser = argparse.ArgumentParser(prog='psij-consol')
+subparser = parser.add_subparsers(dest="command", help='Subcommands')
+validate_parser = subparser.add_parser("validate", help='validate JobSpec file')
+validate_parser.add_argument("file", help="JobSpec file")
+execute_parser = subparser.add_parser("run", help='execute JobSpec file')
+execute_parser.add_argument("file", help="JobSpec file")
```
jameshcorbett, line 19: I wonder if it should accept multiple files? The tricky part would then be how that interacts with ``-n``.

scripts/psij-consol.py
```diff
@@ -0,0 +1,95 @@
...
+execute_parser.add_argument("-j", 
+                            "--job-executor",
```
jameshcorbett, line 21: It looks like the executor needs to be specified so I think this should be a mandatory argument and not an ``--optional``.

scripts/psij-consol.py
```diff
@@ -0,0 +1,95 @@
...
+    if job_spec and  isinstance(job_spec, JobSpec):
+        print("File ok")
+    else:
+        sys.exit("Not a valid file, could not import " + args.file)
+else:
+    
+    if not args.executor:
+        sys.exit("Missing argument executor")
+    
+    print("Importing " + args.file)
```
jameshcorbett, line 71: I think these messages should only be printed if running verbosely.

scripts/psij-consol.py
```diff
@@ -0,0 +1,95 @@
...
+    job_spec = i.load(args.file)
+    if not (job_spec and  isinstance(job_spec, JobSpec)):
+        sys.exit("Something wrong with JobSpec")
+        
+        
+    print("Initializing job executor")
+    jex = psij.JobExecutor.get_instance(args.executor)
+    if not (jex and isinstance(jex, JobExecutor)):
```
jameshcorbett, line 79: Are these checks necessary? Shouldn't the loading mechanism throw a detailed exception in case of failure?

jameshcorbett (commented):
A few nitpicks and questions below.

src/psij/job_spec.py
wilke: Previous implementation did not preserve None value. The check makes sure that the value is propagated and exported correctly as null instead of "None".

scripts/psij-consol.py
wilke: Changed it

scripts/psij-consol.py
wilke: Agreed

scripts/psij-consol.py
wilke: Latest commit changed to print only if verbose.

src/psij/job_spec.py
jameshcorbett: Oh, the string literal ``"None"``? That makes sense.

scripts/psij-consol.py
jameshcorbett: We can always make this change later though.

jameshcorbett (approved):
I would still like to see [this comment](https://github.com/ExaWorks/psi-j-python/pull/162/files#r822156706) addressed but it's minor. Otherwise looks good.

jameshcorbett:
It would be nice to have some basic tests, but that can always be added in another PR.

scripts/psij-consol.py
wilke: It probably does. Changing to try - except.

scripts/psij-consol.py
wilke: Thanks for the clarification. Your suggestion makes the cli more intuitive.

PR #171, opened 2022-03-18
Save envs when testing with cron

hategan:


psij-ci-setup
```diff
@@ -45,74 +48,8 @@ cron_install() {
     echo "$LINE"
     echo "================================================================"
     { crontab -l & echo "$LINE"; } | crontab -
-
```
andre-merzky, line 48: Not part of the change set, but I don't understand the line above:
```sh
crontab -l & echo "$LINE";
```
This runs `crontab -l` in the background.  I assume this is a typo and should be 
```sh
crontab -l && echo "$LINE";
```
?

andre-merzky (approved):
LGTM.  Left one comment on an unchanged line, please check,

psij-ci-setup
hategan: Yeah, I'm trying to understand this, too, and I don't see how it's not a `&&`. I mean, it kinda works, because bash waits for both commands in the list to finish, but that makes `&` even more useless.

andre-merzky: > I mean, it kinda works

Well, only kinda though: it has the chance to discard any existing crontab entries if `crontab -l &` happens to be slower than the `echo ...`.  In my local tests this never happened, but still ... :-)

andre-merzky: Oh, wait - you said bash waits - I wasn't aware of that.  Seems to depend on the pipe consuming command:
```sh
$ date; { (sleep 3 ; date) & true; } | echo; date
Thu Mar 24 00:55:59 CET 2022

Thu Mar 24 00:55:59 CET 2022

$ date; { (sleep 3 ; date) & true; } | cat; date
Thu Mar 24 00:56:04 CET 2022
Thu Mar 24 00:56:07 CET 2022
Thu Mar 24 00:56:07 CET 2022
```

So it works for `crontab -` - I learned something new :-D

hategan: Well, I was trying to figure out why it kinda did work so far :)

andre-merzky (approved):
LGTM

PR #173, opened 2022-03-23
handle errors in status loop of batch schedulers

### LINKED ISSUE #144: Properly deal with errors in the status loop of batch schedulers
If `get_status_command` throws an exception, such as `NotImplementedError`, the `BatchSchedulerJobExecutor` status loop exits and then nothing gets updated any more. This is quite obviously wrong. Exceptions from each job status query should be isolated and reported properly.

andre-merzky:
Fail jobs if `get_status_command` did not yield any usable output, at all, either because the executor was not able to form the correct command or because psij was not able to run it.  In these cases  re-trying would be useless.

This PR though enables a retry if the command did, in fact, yield some output, even if that output may be faulty or incomplete.  We assume here that this could be caused by an intermittent error of the batch system (node overloaded, batch system unresponsive, etc.), and that a retry has a chance of success.

This PR fixes #144.

### DIFF
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -566,6 +566,11 @@ class _QueuePollThread(Thread):
         except subprocess.CalledProcessError as ex:
             out = ex.output
             exit_code = ex.returncode
+        except Exception as ex:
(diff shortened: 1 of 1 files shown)


hategan (approved):
Looks good.

PR #174, opened 2022-03-29
Removed `job._set_status` and integrated that functionality in the

### LINKED ISSUE #48: Move _set_status out of Job
There is a re-usable _set_status method in Job which sets the status and notifies relevant callbacks. This method is not called from Job and is meant to be called from executors. This effectively makes it a public method and reliance on it prevents another possible Job class that correctly implements the spec from working in its place (not that we're encouraging that).

Anyway, _set_status appears to belong more in the JobExecutor abc than in Job, so it should be move there.

hategan:
`job.status` setter.

Renamed `job._executor` into a public property, `job.executor` and moved
the responsibility of setting it to executors (in the `subbmit` and
`attach` methods)

Cleaned up some status setting utility methods in executors.

Fixes #48.

### DIFF
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -181,6 +181,7 @@ class BatchSchedulerExecutor(JobExecutor):
         self._ensure_work_dir()
         assert (job.spec)
 
+        job.executor = self
         context = self._create_script_context(job)
 
         # assumes job ids are unique
@@ -247,6 +248,7 @@ class BatchSchedulerExecutor(JobExecutor):
             The id of the batch scheduler job to attach to.
         """
         job._native_id = native_id
+        job.executor = self
         self._queue_poll_thread.register_job(job)
 
     @abstractmethod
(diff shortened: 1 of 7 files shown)


src/psij/executors/batch/batch_scheduler_executor.py
```diff
@@ -181,6 +181,7 @@ def submit(self, job: Job) -> None:
         self._ensure_work_dir()
         assert (job.spec)
 
+        job.executor = self
```
andre-merzky, line 184: Should we document that an executor implementation SHOULD NOT keep job references around - because otherwise we create circular references and make life difficult for the garbage collector.  

Or we could do something like this?
```py
job.set_executor(self)
class Job():
  ...
  def set_executor(self, executor):
      self._executor_ref = weakref.ref(executor)
  @property
  def executor(self):
      if self._executor_ref:
          return self._executor_ref()
```

Or maybe I am off target and circ refs cannot happen?

andre-merzky (commented):
Nice, looks cleaner now.  Left a question on garbage collection, otherwise looks good to me.

src/psij/executors/batch/batch_scheduler_executor.py
hategan: I think the idea is that executors are more or less permanent objects: you instantiate them at some point and then you keep using them to submit jobs until your "workflow" ends. So my guess is that executors holding references to jobs is a problem with or without circular references.

andre-merzky: That's good enough for me.

andre-merzky (approved):

PR #179, opened 2022-04-04
remove saga executor

### LINKED ISSUE #177: Remove SAGA
Worked as a test case, but likely not needed anymore.

andre-merzky:
This fixes #177


scripts/psijcli.py
```diff
@@ -23,12 +23,11 @@
                                         "flux",
                                         "lsf",
                                         "rp",
-                                        "saga",
```
hategan, line 26: Not really a comment on this PR, but maybe executor names shouldn't be hardcoded.

hategan (approved):
Looks good!

PR #180, opened 2022-04-04
Added support for setting env vars...

hategan:
in some places where it was missing. Also, added a test that checks this.


jameshcorbett (approved):
LGTM. What batch schedulers did you test the implementation on, if any?

hategan:
I didn't, besides the test schedulers. In 24 hrs, we should get reports from the <we_need_a_name_for_it>. But that doesn't cover all schedulers, so manual testing would help.

PR #195, opened 2022-04-27
Docs: API guide

jameshcorbett:
There is still work to do, I just think this is ready to go in and want to keep working on it later.

It is all docstring/docs changes except for one little tweak to ``Job.__str__()``.

### DIFF
--- a/docs/programming.rst
+++ b/docs/programming.rst
@@ -15,12 +15,15 @@ configuration options, to the extent possible.
 
 .. autoclass:: psij.Job
     :members:
+    :noindex:
 
 .. autoclass:: psij.JobStatus
     :members:
+    :noindex:
 
 .. autoclass:: psij.JobState
     :members:
+    :noindex:
 
 Job modifiers
 ^^^^^^^^^^^^^
@@ -36,12 +39,15 @@ scheduling policies.
 
 .. autoclass:: psij.JobSpec
     :members:
+    :noindex:
 
 .. autoclass:: psij.ResourceSpec
     :members:
+    :noindex:
 
 .. autoclass:: psij.JobAttributes
     :members:
+    :noindex:
 
 
 .. _executors:
@@ -87,6 +93,7 @@ The ``psij.JobExecutor`` class is abstract, but offers concrete static methods
 for registering, fetching, and listing subclasses of itself.
 
 .. autoclass:: psij.job_executor.JobExecutor
+    :noindex:
 
 The concrete executor implementations provided by this version of PSI/J Python
 are:
@@ -95,36 +102,43 @@ Cobalt
 ^^^^^^
 
 .. autoclass:: psij.executors.batch.cobalt.CobaltJobExecutor
+    :noindex:
 
 Flux
 ^^^^
 
 .. autoclass:: psij.executors.flux.FluxJobExecutor
+    :noindex:
 
 LSF
 ^^^
 
 .. autoclass:: psij.executors.batch.lsf.LsfJobExecutor
+    :noindex:
 
 PBS
 ^^^
 
 .. autoclass:: psij.executors.batch.pbspro.PBSProJobExecutor
+    :noindex:
 
 Slurm
 ^^^^^
 
 .. autoclass:: psij.executors.batch.slurm.SlurmJobExecutor
+    :noindex:
 
 Local
 ^^^^^
 
 .. autoclass:: psij.executors.local.LocalJobExecutor
+    :noindex:
 
 Radical Pilot
 ^^^^^^^^^^^^^
 
 .. autoclass:: psij.executors.rp.RPJobExecutor
+    :noindex:
 
 .. _launchers:
 
@@ -153,6 +167,7 @@ Like the executor, the ``Launcher`` base class is abstract, but offers
 concrete static methods for registering and fetching subclasses of itself.
 
 .. autoclass:: psij.Launcher
+    :noindex:
 
 The PSI/J Python library comes with a core set of launchers, which are:
 
@@ -161,42 +176,49 @@ aprun
 
 .. autoclass:: psij.launchers.aprun.AprunLauncher
     :members:
+    :noindex:
 
 jsrun
 ^^^^^
 
 .. autoclass:: psij.launchers.jsrun.JsrunLauncher
     :members:
+    :noindex:
 
 srun
 ^^^^
 
 .. autoclass:: psij.launchers.srun.SrunLauncher
     :members:
+    :noindex:
 
 mpirun
 ^^^^^^
 
 .. autoclass:: psij.launchers.mpirun.MPILauncher
     :members:
+    :noindex:
 
 single
 ^^^^^^
 
 .. autoclass:: psij.launchers.single.SingleLauncher
     :members:
+    :noindex:
 
 multiple
 ^^^^^^^^
 
 .. autoclass:: psij.launchers.multiple.MultipleLauncher
     :members:
+    :noindex:
 
 Other Package Contents
 ~~~~~~~~~~~~~~~~~~~~~~
 
 .. automodule:: psij.exceptions
     :members:
+    :noindex:
 
 
 API Reference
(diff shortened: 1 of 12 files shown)


docs/programming.rst
```diff
@@ -1,50 +1,53 @@
...
+``ResourceSpec``, and ``JobAttributes``) are independent of
+executor implementations. The authors strongly recommend that users
+program against these classes, rather than adding executor-specific
+configuration options, to the extent possible.
 
-:class:`~psij.job.Job`
-    Represents a scheduler job, including all its details and state.
+.. autoclass:: psij.Job
+    :members:
+    :noindex:
```
hategan, line 18: The initial intent here was to have an overview of the classes in PSI/J so that users look at one not-so-long-page and get a rough idea of what's there. Then they can click and jump into the gory details. I'm not saying it should be one or the other, but maybe we want to give it some more thought?

docs/programming.rst
```diff
@@ -1,50 +1,53 @@
...
 
-:class:`~psij.resource_spec.ResourceSpec`
+.. autoclass:: psij.JobState
+    :members:
+    :noindex:
 
-    Describes resource requirements for the job, such as number of nodes,
-    etc.
+Job modifiers
```
hategan, line 28: Can we explain the term "modifiers" a bit? Does it refer to the spec + its properties?

docs/programming.rst
```diff
@@ -66,35 +86,63 @@ terminal:
     $ python -m psij plugins
 
 
-The executor implementations provided by this version of PSI/J Python
+JobExecutor Base Class
+^^^^^^^^^^^^^^^^^^^^^^
+
+The ``psij.JobExecutor`` class is abstract, but offers concrete static methods
```
hategan, line 92: Static methods are always concrete.

hategan (commented):
This PR changes the API overview and inlines the class details creating a sort of a single-page API document. I think it is meant to go with a removal of the sphinx auto-generated structure, since that would duplicate the class details, but in the default way that sphinx does it.

There are also some typo fixes.

The second part is a no-brainer. The first presents us with a choice. I think we briefly discussed this choice during the meeting, but it can't hurt for us to take the opportunity to think about the issue in more depth and discuss it on this PR. It would also give this chance to those who were not present at the meeting.

jameshcorbett:
> I think it is meant to go with a removal of the sphinx auto-generated structure, since that would duplicate the class details, but in the default way that sphinx does it.

No, not necessarily. We can have both, and this PR leaves in the auto-generated structure. At the bottom of the API reference page is a link to the complete autogenerated documentation.

docs/programming.rst
jameshcorbett: If you mean that as a general statement about all static methods that's not true, https://stackoverflow.com/a/31590500/10563162

hategan: Ha! I forget we live in pythonworld.

docs/programming.rst
```diff
@@ -1,50 +1,53 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The core API contains the central components of PSI/J. The core API
-classes are independent of executor implementations. User code should
-only program against the core API. The major classes are:
+The Job-related classes listed in this section (``Job``, ``JobSpec``,
+``ResourceSpec``, and ``JobAttributes``) are independent of
+executor implementations. The authors strongly recommend that users
+program against these classes, rather than adding executor-specific
+configuration options, to the extent possible.
```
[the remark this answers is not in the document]
jameshcorbett: What I meant was something like "don't specify a custom option to the Slurm executor for ``--nnodes 15`` when we have ``ResourceSpec`` for that." (I think this maybe answers your above comment too?) I'll add a separate recommendation about executor implementations to the executor section, and clarify the kinds of mistakes I mean in this section.

docs/programming.rst
hategan: > > I think it is meant to go with a removal of the sphinx auto-generated structure, since that would duplicate the class details, but in the default way that sphinx does it.
> 
> No, not necessarily. We can have both, and this PR leaves in the auto-generated structure. At the bottom of the API reference page is a link to the complete autogenerated documentation.

True, that is also an option. I didn't consider it because it duplicates class documentation in multiple places, which I didn't think it's something we want to do. But clearly that's an option that we can discuss.

If we want to be exhaustive, there is also a fourth option :)

hategan: why?!?

hategan:
> > I think it is meant to go with a removal of the sphinx auto-generated structure, since that would duplicate the class details, but in the default way that sphinx does it.
> 
> No, not necessarily. We can have both, and this PR leaves in the auto-generated structure. At the bottom of the API reference page is a link to the complete autogenerated documentation.

True, that is also an option. I didn't consider it because it duplicates class documentation in multiple places, which I didn't think it's something we want to do. But clearly that's an option that we can discuss.

If we want to be exhaustive, there is also a fourth option :)

docs/programming.rst
jameshcorbett: Well, I guess I agree with you if you mean a quick overview of LRMs supported. But if you open up the drop-down bar on the left it lists them all. E.g. click on PSI/J API -> Executors and you get "base class, slurm, lsf" etc. But I also think that the list could go on the front page of the project.

jameshcorbett:
> > > I think it is meant to go with a removal of the sphinx auto-generated structure, since that would duplicate the class details, but in the default way that sphinx does it.
> > 
> > 
> > No, not necessarily. We can have both, and this PR leaves in the auto-generated structure. At the bottom of the API reference page is a link to the complete autogenerated documentation.
> 
> True, that is also an option. I didn't consider it because it duplicates class documentation in multiple places, which I didn't think it's something we want to do. But clearly that's an option that we can discuss.
> 
> If we want to be exhaustive, there is also a fourth option :)

I don't think it's so bad actually, the duplicated class documentation lives on its own page, and it's pretty easy to forget that it exists since the only link to it is within a subsection of a page. And it might come in handy sometime. For example if someone makes a PR adding an executor and they don't update the docs page---well, it will still be in the complete project autogenerated thing.

docs/programming.rst
jameshcorbett: I was just looking for a very generic word to summarize what all the ``JobFoo`` classes have in common. And what I thought was, well, they all change the LRM job's behavior in one way or another.

hategan: Right. Well, they define what the job is and what it needs, which is why we used "Specification". We could change that, but let's try to keep terms to a minimum if we can.

docs/programming.rst
hategan: Not sure what you mean. This is what I get: https://pasteboard.co/5jiFwSl41xOq.png

hategan:
> I don't think it's so bad actually, the duplicated class documentation lives on its own page, and it's pretty easy to forget that it exists since the only link to it is within a subsection of a page.

I'm not sure I understand the above. So let me re-state what I see is happening.
We had:
- an overview page, in the spirit of https://docs.python.org/3/library/, with list of PSI/J classes and a brief description pointing into...
- spinx docs from docstrings, which are exhaustive, one html page per package, and which are meant to be linked into rather than read sequentially
This PR:
- an inlining of the classes in the overview page above, which puts all docs from docstrings for all classes listed there as a one html page (probably not to be read sequentially since it has the same form as the sphinx docs from docstrings documents)
- spinx docs from docstrings, which are exhaustive, one html page per package, and which are meant to be linked into rather than read sequentially

While there are differences between the two items in this PR, the question is whether they are significant. Even if, say, we'd keep both, one might still want an overview page that tells what all the pieces of the API are. A ToC might be enough there, but I'm not sure a useful enough form of it can be generated automatically.

There are other issues, such as, if we link to a class from, say, a tutorial, we'd have to always decide whether we link into one or the other options in this PR, since we now basically have two copies of the docs from docstrings. So we should probably ask ourselves whether and why we would want that.

docs/programming.rst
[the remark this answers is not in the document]
hategan: Ah, I see.

Maybe it's best to call things by their name then. Such as "...rather than setting a custom attribute in JobAttributes".

PR #197, opened 2022-04-29
Docs layout using theme similar to website

rafaelfsilva:


web/index.html
```diff
@@ -1,519 +1,448 @@
...
+                                        contributions to scheduler adapters, testbeds, and specific cluster
+                                        knowledge.
+                                    </p>
+                                </v-card>
+                            </v-col>
+
+                            <v-col justify="center">
+                                <v-card align="center">
+                                    <div class="icon-logo"><i class="fas fa-save" style="color: #FDA214"></i></div>
+                                    <h3><span class="psi-j-font">PSI/J</span> has a rich HPC legacy</h3>
```
jameshcorbett, line 117: Not really a change you made in this PR, but I wonder if "legacy" should be "ancestry". I think "legacy" suggests that things have been built off of it, but really it's that PSI/J is built off of older things.

jameshcorbett (commented):
Big thumbs up for the new colors and figures. If there was a font change, I can't notice it. I do think that the previous organization of the four boxes on the front page was more appropriate though. This PR changes the four boxes to be all on the level with each other, but really the first one is by far the most important---it is the sole reason you would use PSI/J, while the others are all more like strengths to our particular approach.

jameshcorbett (approved):
LGTM, thanks!

PR #202, opened 2022-05-09
Web change main font

hategan:
This PR changes the main font from Gill Sans to something more in-line with the main website (i.e., a more modern display font with multiple weight variants).

It replaces some of the lighter text with a lighter weight, thus maintaining a similar "visual weight", which is what the main site does.

It also makes the docs font follow the main site font.

Once again, this change is open to discussion.

Comparison below:

font-comp

### DIFF
--- a/web/docs/_static/cloud.css
+++ b/web/docs/_static/cloud.css
@@ -11,7 +11,7 @@
 
 @import url("https://cdn.jsdelivr.net/npm/vuetify@2.x/dist/vuetify.css");
 @import url("style.css");
-@import url("theme-3.css");
+@import url("theme-4.css");
 @import url("basic.css");
 
 
@@ -28,7 +28,7 @@ html {
 body {
     padding: 0;
     color: #000;
-    font-family: "Gill Sans", sans-serif;
+    font-family: var(--main-font), sans-serif;
     font-size: 100%;
     line-height: 1.5em;
     margin: 0;
@@ -125,7 +125,7 @@ div.sphinxsidebar h3, div.sphinxsidebar h4 {
     padding: 0;
     margin: 24px 16px 0 0;
 
-    font-family: "Gill Sans", sans-serif;
+    font-family: var(--main-font), sans-serif;
     font-weight: normal;
     color: var(--color-fg-norm-1);
 }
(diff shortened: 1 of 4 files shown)


jameshcorbett (commented):
I have mixed feelings about this, but not strong feelings. For some reason I like the old font better, but I also think I like the main site font better. I also think that the "PSI/J" in "PSI/J runs entirely in user space" and "PSI/J has a rich HPC legacy" needs to be weightier, like before. However, I like that the "Write Scheduler Agnostic HPC Applications" is now darker, just like the text in the three boxes below.

hategan:
> I also think that the "PSI/J" in "PSI/J runs entirely in user space" and "PSI/J has a rich HPC legacy" needs to be weightier, like before.

I fear that I am misunderstanding your statement.

The (new) card headings are 400 and the text is 300. So the heading is weightier. Gill Sans only comes in one weight (at least as far as I can tell), which is 400 (although it looks more like a 500 when you compare it to most other fonts). You can simulate lighter weight by using a lighter shade of gray, but, technically, the card heading and text had the same weight before.

We can try with 500 for headings, but that might be too much (see https://fonts.google.com/specimen/Nunito+Sans). Is that the spirit of your suggestion?

Since Gill Sans is not a google font, you need to have it installed on your system for it to show up, which could be another source of confusion if you are using something else than the image in the first comment to compare things. That said, it's pretty distinctive...

> However, I like that the "Write Scheduler Agnostic HPC Applications" is now darker, just like the text in the three boxes below.

That's in the interest of consistency.

jameshcorbett:
I really meant just the "PSI/J" not the rest of the heading. On my computer it looked like "PSI/J **runs entirely in user space**", like there was special lighter styling for just the "PSI/J"

hategan:
> I really meant just the "PSI/J" not the rest of the heading. On my computer it looked like "PSI/J **runs entirely in user space**", like there was special lighter/thinner styling for just the "PSI/J"

That's funny. Can you post a screenshot?

hategan:
> > I really meant just the "PSI/J" not the rest of the heading. On my computer it looked like "PSI/J **runs entirely in user space**", like there was special lighter/thinner styling for just the "PSI/J"
> 
> That's funny. Can you post a screenshot?

Nevermind. Should be fixed now.

jameshcorbett (approved):
LGTM now thanks

PR #203, opened 2022-05-09
Changed navbar item colors from white to black to improve visibility.

hategan:
This PR changes the navbar items to a dark color for better contrast. It makes the red-ish side of the navbar gradient slightly lighter, for the same reason. 

As usual, this is open to discussion.

nav-color-comp


jameshcorbett (approved):
I think it's fine. Like Kyle I preferred the white, but if W3C says black they probably have a good reason for it.

I think the red at the left end of the navbar could be dulled down (desaturated) a bit.

hategan:
> I think the red at the left end of the navbar could be dulled down (desaturated) a bit.

Yeah, it becomes this pale-pink that's... ugh.
pink

PR #204, opened 2022-05-10
Updated image

hategan:
This PR replaces the image on the main page with something that is a bit more in the style of the first image on exaworks.org.

image-comp


jameshcorbett (approved):
Looks fine. I like Kyle's idea of putting one of the pictures on the docs page, and my preference would be for this new image to go there.

hategan:
Looking again at the two images, the first is more likely to convey the role of PSI/J with respect to a user's needs. So it makes more sense that the first image be kept on the front page. I will close this for now.

PR #207, opened 2022-05-18
Update layout a bit

hategan:
This PR changes the layout of some items on the front page. There were some suggestions that the old layout wasn't quite right. So this one splits the first line into two: the firs is a brief description of PSI/J with an image that illustrates that description. The second is the more detailed main card.

This is one possible layout out of many, so we can use this space to discuss other options.

layout-old

->

layout-new

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -52,18 +52,35 @@
                 <div id="main-banner">
                     <v-container justify="space-around">
                         <v-row justify="center">
-                            <v-col justify="center">
-                                <div>
-                                    <h2>Portable Submission Interface for Jobs</h2>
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
Looks like an improvement to me.

hategan:
> Looks like an improvement to me.

Thanks. Given that @kylechard mentioned some dissatisfaction with the layout, we should wait and see what he has to say.

mturilli:
Maybe I like the more 'dense' one better, less white space

AymenFJA:
Pretty nice and better than the old one. I like the color ranges as well.

kylechard (approved):
Love it.

PR #209, opened 2022-05-18
Change color of selected tab line to something that's visible.

hategan:
selected-tab-old

->

selected-tab-new

### DIFF
--- a/web/theme-4.css
+++ b/web/theme-4.css
@@ -9,7 +9,7 @@
     --color-bg-norm-1: #ffffff;
     --color-bg-norm-2: #ffffff;
     --color-link: #6080ff;
-    --color-fg-lines: #fe7000;
+    --color-fg-lines: #6b6bff;
     --color-global-bg: #ffffff;
     --main-font: "Nunito Sans" !important;
 }


jameshcorbett:
Fine by me but could we make it a deep brick red or similar rather than blue?

hategan:
> Fine by me but could we make it a deep brick red or similar rather than blue?

No? It's an accent color. Intentionally picked on the opposite side of the color wheel from the other ones, which are relatively clustered in the red-orange. I mean, it doesn't have to be that, but that's how you usually pick accent colors.

jameshcorbett:
Wait now I am confused. This is something that's supposed to show up when you hover over a menu button? If so I don't really see a need since when I go to the website there's a pretty obvious change in the button when I hover over it.

jameshcorbett (approved):
I suggested deep brick red (as in, so deep it doesn't look red any more) because I thought the bright orange / bright blue combination was kind of horrible. There are 3 dimensions to create contrast, and I was suggesting to do it along the value and saturation dimensions rather than hue. Now that I pull it up in my browser it actually looks fine (TBH it's not really even noticeable) so I'm fine with this going in but I still think it should be darkened or desaturated or both.

hategan:
> bright orange / bright blue combination was kind of horrible.

I don't think one can objectively make statements about the beauty of color combinations, or we wouldn't much visual art. I clearly thought this one reasonable, or it wouldn't be there.

Most places you look at will tell you that you pick the accent *color* to, ehm, accentuate. And the best way to do that *with color* is on the other side of the wheel (although you could have other palettes, that make more creative use of the wheel -- see, e.g., https://paletton.com/). You could say that that's a horrible idea, but I don't think you would have the support of most design resources out there. That's not to say that that's necessarily the best choice in every case.

So anyway, we can try things. You can fiddle with it yourself for a bit. It's `--color-fg-lines` in `theme-4.css`. You can put the rgb of what you mean by deep brick red there. If on linux, use `./web/watch.sh` instead of `build.sh` to automatically compile your edits. Then, if you find a color you like, take a screenshot and post it here and we can discuss.

jameshcorbett:
Screen Shot 2022-05-19 at 9 23 22 AM

The color is #612f2f. It looks black against the navbar but it's actually quite red.

hategan:
> Screen Shot 2022-05-19 at 9 23 22 AM
> 
> The color is #612f2f. It looks black against the navbar but it's actually quite red.

selected-tab-new

I guess we disagree on what's visually more pleasing. Let's have the others weigh in.

hategan:
> I guess we disagree on what's visually more pleasing. Let's have the others weigh in.

Actually, no. There's ample precedent in design for using blue in this case, this is a minor issue that's not worth wasting time on while there's plenty of other work to be done, so I'm just going to merge it. If it turns out to be terribly problematic, we can change colors later.

jameshcorbett:
That's fine by me. It's never going to turn out to be terribly problematic 😜

hategan:
> That's fine by me. It's never going to turn out to be terribly problematic

I doubt folks will even notice. That said, it's also used as the color of the divider in the docs, so maybe somebody will!

PR #212, opened 2022-06-06
Fix for when the PSI/J site is not at the root of the web path.

hategan:
Docs are currently broken because the JS code didn't account for the fact that the current site does is not at the root at the domain. This PR fixes that.

### DIFF
--- a/web/lib.js
+++ b/web/lib.js
@@ -7,7 +7,6 @@ docsHashChange = function(path, hash) {
     else {
         path = path.substr(ix);
     }
-    console.log("Path: " + path);
     if (hash.startsWith("#")) {
         hash = hash.substr(1);
     }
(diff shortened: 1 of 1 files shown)


web/lib.js
```diff
@@ -1,8 +1,13 @@
...
-        "(should start) with '/docs/'" + path);
+    var ix = path.indexOf("/docs/");
+    if (ix == -1) {
+        throw new Error("Something is wrong with the docs path " +
+        "(should start) with '/<prefix>/docs/'" + path);
     }
+    else {
+        path = path.substr(ix);
+    }
+    console.log("Path: " + path);
```
andre-merzky, line 10: Do you want to leave this log in?

andre-merzky (approved):
This looks reasonable to me - thanks!

web/lib.js
hategan: No. Thanks for catching that.

PR #214, opened 2022-06-06
The list looked weird and more like part of the ToC than the paragrap…

hategan:
…h it seemingly belonged to.


andre-merzky (approved):
LGTM.

PR #217, opened 2022-06-15
PBS attributes are all short.

### LINKED ISSUE #211: PBS executor and custom attributes
The PBS executor translates custom attributes into directives of the form `#PBS --{{key}}={{value}}`. However, PBS uses short options almost exclusively (except `--version` -- see https://help.altair.com/2021.1.3/PBS%20Professional/PBSReferenceGuide2021.1.3.pdf).

So things should be updated so that custom attributes are translated to `#PBS -{{key}}={{value}}`.

hategan:
Fixes #211


andre-merzky (approved):
LGTM

PR #219, opened 2022-06-21
More tests

hategan:
This PR started as an effort to add more tests, particularly in the direction of resource specification/allocation.

It needed some local validation, so a test launcher was added that can fake multiple hosts.

As a side effect, ResourceSpecV1 was also revamped a bit: it made no sense that one couldn't specify, say `process_count=4` and `node_count=2`. That is now possible. ResourceSpecV1 checks basic constraints, such as `process_count % node_count = 0`, `process_count = node_count * ppn`. It can also infer some of these numbers if they are not otherwise specified. However, whether the inferred numbers are used or not by an executor is left to the executor.

### DIFF
--- a/src/psij/resource_spec.py
+++ b/src/psij/resource_spec.py
@@ -4,14 +4,6 @@ from typing import Optional, List
 from psij.exceptions import InvalidJobException
 
 
-def _nulls(objs: List[object]) -> int:
-    s = 0
-    for e in objs:
-        if e is None:
-            s += 1
-    return s
-
-
 class ResourceSpec(ABC):
     """
     A base class for resource specifications.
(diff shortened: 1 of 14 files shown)


hategan:
Sorry. Converted the wrong PR to draft.

src/psij/resource_spec.py
```diff
@@ -1,9 +1,17 @@
...
 from psij.exceptions import InvalidJobException
 
 
+def _nulls(objs: List[object]) -> int:
+    s = 0
+    for e in objs:
+        if e is None:
+            s += 1
+    return s
+
```
andre-merzky, line 13: Isn't this the same as `list.count(value)`?
```py
>>> l = [1, None, 2, None, 3]
>>> l.count(None)
2
```

andre-merzky (approved):
LGTM.  Left a comment unrelated to the added tests.

src/psij/resource_spec.py
hategan: Silly me. Yes. Will update.

PR #220, opened 2022-06-28
Use per-process deployment directories for launcher files

hategan:
Unfortunately, a single directory can lead to clashes between versions and processes.

### DIFF
--- a/src/psij/launchers/script_based_launcher.py
+++ b/src/psij/launchers/script_based_launcher.py
@@ -68,6 +68,8 @@ class _ProcessSandbox:
 
     def clean(self) -> None:
         now = time.time()
+        if not self.config.work_directory.exists():
+            return
         for child in self.config.work_directory.iterdir():
             if child.name.startswith(_SANDBOX_PREFIX):
                 marker = child / '.stale'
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
Sandbox dir's make sense IMHO - it also avoids the state dir filling up too quickly.  But if I read the code correctly, the sandboxes are marked scale via an `atexit` hook. Yes, there is that 2-day timeout, but I would still worry that we collect sandboxes for jobs which are still running or where the sandbox is still needed to collect job state even when the job completed some days before.  What I am basically trying to say is that, IMHO, the timeout should be more like a month to play things safe, and this should be clearly documented so that people don't get surprised that the sandboxes disappear.

Otherwise this looks all good to me.

hategan:
> the timeout should be more like a month to play things safe

Yes, I think I agree. The exit handler does not deal with jobs that will be re-attached. Perhaps we can implement a simple in-process counter for an even better solution.

An alternative that I was thinking about is checksum based naming in which we have a single directory, but files are named by their checksum (done with some sufficiently strong algorithm). That makes the management a bit easier. So I might retract this PR.

hategan:
I'm changing this to a draft. I don't expect it to be merged per our earlier discussion. An alternative solution is being tested. However, until and if that alternative is adopted, I think this should remain open for comparison purposes.

PR #221, opened 2022-06-28
Not all mpiruns recognize --oversubscribe. OpenMPI does, so check.

hategan:


src/psij/launchers/scripts/mpi_launch.sh
```diff
@@ -5,10 +5,19 @@ source $(dirname "$0")/launcher_lib.sh
 _PSI_J_PROCESS_COUNT="$1"
 shift
 
+IS_OPENMPI=0
+if mpirun -version | grep "Open MZI" >/dev/null 2>&1; then
```
andre-merzky, line 9: There is a typo (`MZI`)

src/psij/launchers/scripts/mpi_launch.sh
```diff
@@ -5,10 +5,19 @@ source $(dirname "$0")/launcher_lib.sh
...
+
 pre_launch
 
 set +e
-mpirun -n $_PSI_J_PROCESS_COUNT "$@" 1>$_PSI_J_STDOUT 2>$_PSI_J_STDERR <$_PSI_J_STDIN
+if "$IS_OPENMPI" == "1"; then
+    mpirun --oversubscribe -n $_PSI_J_PROCESS_COUNT "$@" 1>$_PSI_J_STDOUT 2>$_PSI_J_STDERR <$_PSI_J_STDIN
+else
+    mpirun -n $_PSI_J_PROCESS_COUNT "$@" 1>$_PSI_J_STDOUT 2>$_PSI_J_STDERR <$_PSI_J_STDIN
+fi
```
andre-merzky, line 20: Once the number of cases increase (and they will), we should reconsider the code replication.  For now that works.

andre-merzky (approved):
Marked typo, otherwise LGTM - thanks.

src/psij/launchers/scripts/mpi_launch.sh
hategan: Uh, that's not a typo. I was testing both possibilities (string found and string not found) and forgot to revert. Will fix.

PR #222, opened 2022-06-28
Alternate file deployment scheme 2

hategan:
This is an alternative to #220. The problem with #220 was that each new process would create a new directory, but there was no reasonable way to clean these directories. Cleaning too frequently could interfere with debugging and would be incompatible with long-running detached jobs. Setting arbitrary ages, such as 2 days or one month is equally distasteful.

This PR changes launcher so that each launcher instance deploys all its files at once, since these files may reference one another. It does the deployment in directories named as checksums of the contents of the files within. Changing one of the respective files in the source would result in new directories and, more importantly, unchanged launcher files would not. This minimizes clutter enough that manual cleaning is feasible if at all necessary. 

Note: this PR is on top of the more_tests branch, so it looks big. Compare accordingly when reviewing.

### DIFF
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -504,6 +504,7 @@ class BatchSchedulerExecutor(JobExecutor):
                     # only read output from submit script if another error message is not
                     # already present
                     status.message = self._read_aux_file(job, '.out')
+                    logger.debug('Output from launcher: %s', status.message)
                 else:
                     self._delete_aux_file(job, '.out')
 
@@ -525,6 +526,7 @@ class BatchSchedulerExecutor(JobExecutor):
             finally:
                 self._delete_aux_file(job=job, suffix=suffix, path=path, force=True)
         else:
+            logger.debug('%s does not exist', path)
             return None
 
     def _delete_aux_file(self, job: Optional[Job] = None, suffix: Optional[str] = None,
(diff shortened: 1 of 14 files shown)


andre-merzky (commented):
Yeah, the diff is large because of the other PRs.  If you don't mind I'd rather review when those are merged.

Meanwhile though: I don't get the rationale for the hash, really: what advantage does it have over using the job ID?  It seems it only makes things harder to debug and harder to reproduce.  And the cleanup problem should stay the same I think - or am I missing something?

hategan:
> Yeah, the diff is large because of the other PRs. If you don't mind I'd rather review when those are merged.

Yeah, it was supposed to be on top of more_tests. Not sure how to make it show the diff against that.

> 
> Meanwhile though: I don't get the rationale for the hash, really: what advantage does it have over using the job ID?

Once per unique set of launcher files vs. once per job. The first should only generate one dir for a given launcher and a given (unmodified) version of psi/j. There is no cleanup, since we don't expect this list to grow too large.

If we do per job, we're back to the "when do we clean" problem (i.e., 2 hours, 2 days, when process ends, how about jobs we plan to re-attach to, etc.)

andre-merzky (approved):
Oh, of course - the assumption is that the launcher script is independent of the job itself and thus remains unchanged between job invocations...  Well, fair, as long as this is the case this works.  I am still not a big fan of hash-generated directory and file names, but I also don't see anything wrong with it really: it solves the problem at hand.  Thanks!

hategan:
> Yeah, the diff is large because of the other PRs. If you don't mind I'd rather review when those are merged.

@andre-merzky, should be more manageable now

PR #225, opened 2022-07-21
Adding MIT license for release

yadudoc:
Adding MIT License based on discussion at Exaworks workshop. We need a license to be on for pushing out a Pypi release.

With agreement from: @kylechard @danlaney @shantenujha @hategan and @rafaelfsilva

PR #226, opened 2022-07-21
Release prep updates

yadudoc:
This PR contains a bunch of changes required for releasing over Pypi:

* New `tag_and_release.sh` script to tag and release over pypi
* Updates to Makefile to call the above script. Do something like this `make VERSION="<VERSION_STRING>" tag-and-release
* psij version is now defined in `src/psij/version.py`
* bug fix to `setup.py` to actually enumerate `requirements.txt` as `install_requires`
* a few optional extra requirements files are now deleted and replaced as `optional_extras` in `setup.py`
* setup.py now specifies the MIT Licence that was approved in #225


yadudoc:
Looks like the readthedocs build is now failing, since it tries to install `requirements-doc.txt` which no longer exists. The fix would be to update the build options at rtd to install as `pip install .[docs]`.

hategan:
> Looks like the readthedocs build is now failing, since it tries to install `requirements-doc.txt` which no longer exists. The fix would be to update the build options at rtd to install as `pip install .[docs]`.

That's not the same though, right? That installs what was in requirements-docs.txt as well as psij from the current dir.

hategan (changes requested):
The PR adds some tools to help make a release to pypi. It also moves some of the requirements to setup.py extras. 

Moving the requirements to extras seems to make sense in the case of executors with dependencies that are not installed by default, such as saga.

However, this makes a bit less sense for requirements-dev.txt and requirements-docs.txt, since they are not extras, but dependencies specific to development tools and documentation tools needed to develop psij or build documentation, which are processes that are separate from using an installed psij.

In other words, if one wants to hack at psij, there are many ways to do so not all of which involve `pip install -e .`. I might also want to build the documentation, which shouldn't involve installing psij locally.

So let's keep r-dev.txt and r-docs.txt.

yadudoc:
Marking this as draft, since there are a bunch of conflicts. I'll split the changes out to smaller PRs. Once those are in, I'll close this.

hategan:
Can we close this now?

PR #230, opened 2022-07-28
Setup version info

yadudoc:
This PR splits out changes in #226 to support better version info handling.
 
* version info is now set in the module `src/psij/version.py`
* updates to `setup.py` to use the above
* `psij.__version__` will work now.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -60,4 +60,6 @@ jobs:
           pip install -r requirements-connector-radical.txt
           make launcher-scripts
           make install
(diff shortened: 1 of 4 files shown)


hategan (approved):
Looks good.

andre-merzky (approved):
LGTM, thanks.

yadudoc:
pytest is failing due to issues with figuring out the git branch. There's a bunch of code here that tries to figure out the branch info from config, env and some calls to git cmdline that I do not want to touch. For github actions, it might be simpler to rely on the env var `GITHUB_REF_NAME`.

yadudoc:
Merging now that all tests are green.

hategan:
> pytest is failing due to issues with figuring out the git branch.

Not sure about that. The error is in teardown. The branch is figured out in setup. There's a few other things in a try block before the git branch is determined. If any of those fails, control won't reach the git branch determination part, which I suspect is what might be happening.

PR #231, opened 2022-07-28
Adding deploy scripts

yadudoc:
Adding deploy scripts and matching changes to the `Makefile`.

Here are the deploy steps I recommend:
1. Create a branch off of main once ready to deploy
2. Bump the version info, docs etc in the branch
3. Review and merge to main
4. Run the following to deploy from main 
```
make VERSION=<version_info> tag-and-release
```

The above call will make a github tag, and deploy to pypi, so make sure to have your pypi creds handy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,3 +60,10 @@ install:
 .PHONY: develop
 develop:
 	$(PYTHON) setup.py develop
+
+
+.PHONY: tag-and-release
+tag: ## create a tag in git. to run, do a 'make VERSION="version string" tag-and-release
+	./tag_and_release.sh create_tag $(VERSION)
+	./tag_and_release.sh package
+	./tag_and_release.sh release
--- a/tag_and_release.sh
+++ b/tag_and_release.sh
@@ -1,0 +1,54 @@
+#!/bin/bash
+set -euf -o pipefail
+
+create_tag () {
+    if [ -z "$1" ]
+      then
+        VERSION="unknown"
+      else
+        VERSION=$1
+    fi
+
+    PSIJ_VERSION=$(python3 -c "import psij; print(psij.__version__)")
+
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -60,3 +60,10 @@ install:
 .PHONY: develop
 develop:
 	$(PYTHON) setup.py develop
+
+
+.PHONY: tag-and-release
+tag: ## create a tag in git. to run, do a 'make VERSION="version string" tag-and-release
```
hategan, line 66: "tag" vs "tag-and-release"?

Looks good otherwise. I'm assuming this is based on what's in Parsl.

yadudoc: Named it that way, because it does both tag followed by release. Are you recommending that I break it up into two separate entry points?

wilke: We might if we need to tag separately of releasing. I think it is ok for now unless you have different use cases in mind.

wilke (approved):
Looks good. Thanks and tag-and-release is fine.

Makefile
andre-merzky: The `PHONY` target is named `tag-and-release`, the make target is named `tag` - shouldn't that be the same?

yadudoc: Yep, that's a bug. Let me fix that in a different PR since this one got merged.

andre-merzky:
Sorry for not checking earlier, but `tag_and_release.sh` is missing the executable bit.  Either that bit should be set, or the makefile should explicitly call `/bin/bash tag_and_release.sh` I think?

yadudoc:
@andre-merzky Thanks for catching that. It's odd that it worked on my side without the x bit. I've made the fixes on #232

Makefile
hategan: It may be useful to address comments before merging in the future :)

PR #235, opened 2022-07-31
Fail early if git info cannot be obtained and saving or uploading is …

hategan:
…enabled.

Otherwise print a warning and continue.

This also disables the building of the data package for upload or save in test suite teardown if neither saving or uploading are enabled.

The basic issue with getting git info in gh actions is that using checkout/v2 checks out a particular hash with a detached head. That causes most complex git stuff to fail. However, said complex git stuff is only used to collect data for saving or uploading and not when running tests otherwise.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -45,6 +45,7 @@ jobs:
         run: |
           cd /workspace
           git --version
+          git branch -a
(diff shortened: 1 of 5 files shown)


andre-merzky (approved):
Looks reasonable to me.

PR #237, opened 2022-08-02
Minimal test uploads

hategan:
This PR adds the possibility of uploading minimal test results. This is best described by the corresponding text in `testing.conf`:

> Enables minimal uploads mode. In minimal mode, no information that is
> potentially sensitive is uploaded to the test results server. Specifically,
> only the following are uploaded:
>   - code related identifiers, such as branch names, class names, etc.
>   - times and dates
>   - test related identifiers, such as run id
>   - the maintainer email, if one is provided
>   - boolean values, such as whether the target system has mpirun available,
>     etc.
>   - git identifiers and counts: branch name, last commit hash, how many
>     commits the local code has ahead/behind of the remote, and if there
>     are any local changes
>   - the result status of tests, excluding string error messages coming from
>     tests
>
> Excluded items include, but are not limited to:
>   - stdout/stderr from tests
>   - logs from tests
>   - filesystem paths
>
> The full information is still saved locally and can be accessed and manually
> vetted if required. Setting this to "true" automatically enables saving of
> results to files.

What is included in the upload is defined by a whitelist (specified above). That is, nothing outside the whitelist is included.

### DIFF
--- a/testing.conf
+++ b/testing.conf
@@ -68,6 +68,9 @@ scope = core
 #   - the maintainer email, if one is provided
 #   - boolean values, such as whether the target system has mpirun available,
 #     etc.
+#   - git identifiers and counts: branch name, last commit hash, how many 
+#     commits the local code has ahead/behind of the remote, and if there
+#     are any local changes
 #   - the result status of tests, excluding string error messages coming from
 #     tests
 #
--- a/testing.conf
+++ b/testing.conf
@@ -81,7 +81,8 @@ scope = core
 #
 # The full information is still saved locally and can be accessed and manually
 # vetted if required. Setting this to "true" automatically enables saving of
-# results to files.
+# results to files. Please note that this option modifies the behavior of 
+# uploads, which must still be requested using the "--upload-results" flag.
 #
 # minimal_uploads = false | true
 
--- a/testing.conf
+++ b/testing.conf
@@ -59,6 +59,33 @@ scope = core
 
 
 #
+# Enables minimal uploads mode. In minimal mode, no information that is 
+# potentially sensitive is uploaded to the test results server. Specifically, 
+# only the following are uploaded:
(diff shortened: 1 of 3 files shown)


testing.conf
```diff
@@ -58,6 +58,36 @@ maintainer_email =
...
+#     tests
+#
+# Excluded items include, but are not limited to:
+#   - stdout/stderr from tests
+#   - logs from tests
+#   - filesystem paths
+#
+# The full information is still saved locally and can be accessed and manually
+# vetted if required. Setting this to "true" automatically enables saving of
+# results to files.
```
andre-merzky, line 84: The code seems to only save the results if `upload` is also set (line 465 in `conftest.py`), so the above statement is only half-true.  I would be in favor of following the documentation above and use 
```py
if minimal:
        save = True
```
but clarifying the documentation also works.

andre-merzky (approved):
The 4 methods added to `conftest.py` would benefit from a simple unit test - let me write those (#238).  This PR can be merged meanwhile IMHO.

testing.conf
hategan: Uploads and saves are enabled by passing special flags (--upload-results and --save-results) to pytest and not through settings in testing.conf.
minimal_uploads was meant to change how --upload-results behaves.

Wouldn't it be better for consistency to update the docs?

hategan:
> The 4 methods added to `conftest.py` would benefit from a simple unit test - let me write those (#238).

Indeed. Thank you.

testing.conf
andre-merzky: > Wouldn't it be better for consistency to update the docs?

Yes, I agree.

hategan: Thanks. Let me know if the new wording sounds reasonable.

hategan:
> Are we ready to merge seeing that this was approved?

Yes. Sorry for the delay.

PR #244, opened 2022-08-24
Adding log level config from env

### LINKED ISSUE #243: Enable logging globally and in production
I would like to enable logging and be able to set log level in a production environment. The current setup hs logging enabled but logs only default level. I suggest reading the log level from the environment at initialization time.

```
if os.environ['PSIJ_LOG_LEVEL']:
    logging.basicConfig(format='%(name)s:%(levelname)s:%(message)s',
                        level=os.environ['PSIJ_LOG_LEVEL'].upper())
logger = logging.getLogger(__name__)
```

wilke:
closes #243

### DIFF
--- a/src/psij/__init__.py
+++ b/src/psij/__init__.py
@@ -26,7 +26,11 @@ __all__ = [
     'InvalidJobException', 'UnreachableStateException', 'Export', 'Import'
 ]
 
+if os.environ['PSIJ_LOG_LEVEL']:
+    logging.basicConfig(format='%(name)s:%(levelname)s:%(message)s',
+                        level=os.environ['PSIJ_LOG_LEVEL'].upper())
 logger = logging.getLogger(__name__)
+
 
 T = TypeVar('T')
 
--- a/src/psij/__init__.py
+++ b/src/psij/__init__.py
@@ -26,7 +26,7 @@ __all__ = [
     'InvalidJobException', 'UnreachableStateException', 'Export', 'Import'
 ]
 
-if os.environ['PSIJ_LOG_LEVEL']:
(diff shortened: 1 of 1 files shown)


src/psij/__init__.py
```diff
@@ -26,8 +26,12 @@
     'InvalidJobException', 'UnreachableStateException', 'Export', 'Import'
 ]
 
+if os.environ['PSIJ_LOG_LEVEL']:
```
hategan, line 29: If PSIJ_LOG_LEVEL is not defined, that line will throw an exception and prevent any code using psij from working.

I think what you want to check is `if 'PSIJ_LOG_LEVEL' in os.environ`:

wilke: Changed, thanks for the suggestion

hategan:
Do we maybe want to provide a more comprehensive solution, such as `PSIJ_LOG_CONFIG` that points to a file which would allow users to configure all logging aspects?

That said, PSI/J is a library. It is not meant to be run in a standalone fashion. In particular, since logging configuration is global, any PSIJ_ logging configuration options would, in fact, apply to all python code that uses psij and may lead to conflicts if user code configures different options. 

We should perhaps discuss this, but it seems like logging configuration should be left to the main application that uses PSI/J.

kylechard:
Discussed on call and agreed this should be moved to the CLI.

hategan:
Moving to 1.0.0

hategan:
Changing this to draft until it gets moved to the CLI and it's ready for review

hategan:
I'm closing this since there has been no activity on it for over 6 months.

PR #246, opened 2022-08-25
Adding quickstart draft

kylechard:


hategan:
You need a valid .rst. If you click on "Details" for the failing checks, you'll see:

    /home/runner/work/psi-j-python/psi-j-python/docs/quickstart.rst:17: WARNING: Title underline too short.

    Install from source
    ^^^^^^^^^^^^^^^^

You need to add three "^"s.

The second issue is

    /home/runner/work/psi-j-python/psi-j-python/docs/quickstart.rst: WARNING: document isn't included in any toctree

See here for a solution: https://stackoverflow.com/questions/15249340/warning-document-isnt-included-in-any-toctree-for-included-file
In other words, either add the doc to the toctree in index.rst or add is as an orphan if we only want it linked externally and not from the index.

PR #249, opened 2022-08-30
When a doc page (that was meant to be loaded as an iframe inside the …

### LINKED ISSUE #248: Deal with duplicate docs pages on web site
The docs pages are served by the main site in an iframe and integrate with the rest of the web site. However, they are still accessible directly. When accessed directly, they are not fully functional. For example:

[https://exaworks.org/psi-j-python/docs/tutorial_add_executor.html](https://exaworks.org/psi-j-python/docs/tutorial_add_executor.html) is the raw iframe content

and 

[https://exaworks.org/psi-j-python/#docs/tutorial_add_executor.html](https://exaworks.org/psi-j-python/#docs/tutorial_add_executor.html) is the same document integrated in the PSI/J site.

While the former isn't linked directly, it may still be indexed and linked to by search engines. We should make it such that a user is always redirected to a functional version of documentation pages.

hategan:
…main page) is loaded directly, detect that and redirect to the proper one.

Fixes #248.

### DIFF
--- a/web/docs/_static/cloud.js
+++ b/web/docs/_static/cloud.js
@@ -993,7 +993,14 @@
         parent.docsHashChange(window.location.pathname, window.location.hash)
     });
     $(window).on("load", function() {
-        parent.docsLoaded(window.location.pathname, window.location.hash)
+        if (window.parent.location == window.location) {
+            // not in an iframe; redirect
+            window.location.href = "/#" + window.location.pathname.substring(1);
+        }
+        else {
+            // in an iframe; not necessarily in the main page, but that is an unlikely scenario
+            parent.docsLoaded(window.location.pathname, window.location.hash);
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
The intent if his change sounds reasonable to me.  But I don't know much js, would be good if someone else could have another look at this...

hategan:
Sorry for not mentioning this earlier. I have tested this locally. I think it's somewhat difficult to ensure (or even define) correctness when it comes to web technologies, but it is 1) an improvement and 2) targeted to a very specific case (window.parent == window -- which, according to https://developer.mozilla.org/en-US/docs/Web/API/Window/parent should never throw an exception) that is broken as is. So I think it's reasonable to deploy this and see if it behaves.

PR #258, opened 2022-09-01
uniform cb registration

### LINKED ISSUE #247: API inconsistent for status callbacks
Setting a status callback on a executor:
```py
def state_cb(job, status):
    print(job.id, str(status)
    
jex = psij.JobExecutor.get_instance(...)
jex.set_job_status_callback(cb)
```

Setting a status callback on a job:
```py
class StateCB(JobStatusCallback):
    def job_status_changed(self, job, status):
        print('Job {}: {}'.format(job.id, status))

job = Job(JobSpec(executable='/bin/date'))
job.status_callback = StateCB()
```

Two questions:
  - 1) should we make this uniform?
  - 2) which way?
  
My opinion:
  - 1) yes
  - 2) use the executor approach also for jobs

andre-merzky:
This PR fixes #247 in that it makes the registration of job status callbacks uniform across `psij.Job` and `psi.JobExecutor` instances, mostly by lifting code from the executor.

### DIFF
--- a/src/psij/job.py
+++ b/src/psij/job.py
@@ -2,7 +2,7 @@ import logging
 import threading
 from abc import ABC, abstractmethod
 from datetime import timedelta, datetime
-from typing import Optional, Sequence
+from typing import Optional, Sequence, Union, Callable
 from uuid import uuid4
 
 import psij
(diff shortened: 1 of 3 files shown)


hategan (approved):
Thanks. Looks good.

PR #259, opened 2022-09-01
make mandatory args not-optional, raise `TypeError` on type errors

### LINKED ISSUE #240: Serializer should raise exception, not call `sys.exit()`
[Example](https://github.com/ExaWorks/psi-j-python/blob/main/src/psij/serialize.py#L44)

Took me quite a while to understand why my code just exited randomly w/o triggering my fail safeties.... :-P  Or maybe I am missing the point behind the `sys.exit` calls?

andre-merzky:
This fixes #240.

### DIFF
--- a/src/psij/serialize.py
+++ b/src/psij/serialize.py
@@ -39,7 +39,6 @@ class Export(object):
 
     def export(self, obj: object, dest: str) -> bool:
         """Serializes an object to a file."""
-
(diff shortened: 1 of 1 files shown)


jameshcorbett (approved):
LGTM, thanks Andre!

wilke (commented):
Looks good. Thanks.

PR #275, opened 2022-09-08
Added selectors for docs

hategan:
This PR adds selectors for executors (and more) to the docs. It ended up being a pure JS solution rather than a Sphinx extension. See the [README](https://github.com/ExaWorks/psi-j-python/blob/add_selectors_to_docs/docs/README-selectors.md) for details on how to use and see image below for an idea of how this looks like.
doc-selectors


andre-merzky (approved):
This looks neat and useful!  I can't comment on the javascript part though.

Two questions:
  - an empty file `docs/_static/extras.css` was committed - on purpose or by accident?
  - that drop-down seems useful for other things than executor type, like host, task types, etc.  Would it make sense to generalize `executor-type-selector` to `type-selector` etc?

hategan:
> * an empty file `docs/_static/extras.css` was committed - on purpose or by accident?

I had it there with stuff, but then it seemed like said stuff belonged better in the web/docs/theme css. But then it felt like maybe there should still be things in extras.css, except there aren't. I should probably delete it.

> * that drop-down seems useful for other things than executor type, like host, task types, etc.  Would it make sense to generalize `executor-type-selector` to `type-selector` etc?

It is generic, although you have to explicitly add the new types. See the bottom of the readme. The reason why we don't use one class for all of them is because they are synchronized (when I change one selector, all selectors change, similar to the tabs in the spec). But we have to synchronize executors separately from, say, hosts, task types, etc.

andre-merzky:
> It is generic, although you have to explicitly add the new types. See the bottom of the readme. The reason why we don't use one class for all of them is because they are synchronized (when I change one selector, all selectors change, similar to the tabs in the spec). But we have to synchronize executors separately from, say, hosts, task types, etc.

Thanks for the reply, I missed that indeed.  That makes sense then as is.

Issue #283, opened 2022-09-15
Deploy test on theta


wilke:
Setup psij testing suite to run daily, either as cron job or via gitlab.

wilke:
Status: **Failing**

Steps:

1. loading python3.9
2. creating venv 
3. installing requirements.txt and requirements-tests.txt
4. ./psij-ci-run

ModuleNotFoundError: No module named 'requests'

wilke:
Status: **Failing**

manually installed: python -m pip install requests
./psij-ci-run: 

./psij-ci-run
From https://github.com/ExaWorks/psi-j-python
 * branch            HEAD       -> FETCH_HEAD
Reading configuration... OK
Cloning repository... OK
Testing branch "main"... psij.venv/bin/python: No module named pytest
OK
Testing branch "combine-getting-started-quickstart-pages"... psij.venv/bin/python: No module named pytest
OK
Testing branch "wilke/logging"... psij.venv/bin/python: No module named pytest
OK
Testing branch "1.0.0-a0_release_branch"... FAILED

hategan:
Did you follow the official instructions (README-testing.md)?

If you did, it would probably have told you to either run `psij-ci-setup`, which installs all needed dependencies, or to use `requirements-dev.txt` which does the same. Can you try that and submit an issue if *that* fails?

hategan:
PS: `requests` is needed by Sphinx, which is listed in `requirements-docs.txt`, which is included from `requirements-dev.txt`.

It should probably be listed explicitly.

wilke:
Seems to work but fails with the error below. Probably a problem in the branch but need to check.

```
Testing branch "wilke/logging"... ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --minimal-uploads
  inifile: None
  rootdir: /tmp/tmpjxup64ta/code

OK
Testing branch "1.0.0-a0_release_branch"... FAILED
Traceback (most recent call last):
  File "/lus/theta-fs0/projects/CSC249ADTR01/psi-j-python/tests/ci_runner.py", line 195, in <module>
    run_tests(conf, site_ids, branches, clone)
  File "/lus/theta-fs0/projects/CSC249ADTR01/psi-j-python/tests/ci_runner.py", line 156, in run_tests
    run_branch_tests(conf, tmpp, run_id, clone, site_id, branch)
[...]

PR #284, opened 2022-09-15
Make req-dev include req-tests and add requests explicitly to req-tests

hategan:
This should address #283.

@wilke, can you try setting up the tests with this branch?

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,10 +5,9 @@
 -r requirements-connector-radical.txt
 
 # Testing / QA
+-r requirements-tests.txt
 mypy >=0.931
-pytest >= 6.2.0
 flake8
 flake8-docstrings
 autopep8
 types-requests
-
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
Looks reasonable to me, FWIW.

wilke (approved):

wilke:
Command: ./psij-ci-run
Status: **OK**

Tests are running and i installed the requirements with `python -m pip install --target .packages --upgrade -r requirements-dev.txt`

There is one error message at build time:

```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

sphinx-tabs 3.4.1 requires docutils~=0.18.0, but you'll have docutils 0.17.1 which is incompatible.
```

PR #287, opened 2022-09-15
docs: reorganize developer docs

jameshcorbett:
I think this might be a modest improvement in organization.


jameshcorbett:
Just to clarify the main change is that the architecture, "adding an executor tutorial", and contributing guide pages are given their own folder. From a user's perspective, they are now hidden under one layer of indirection. The side bar reads "getting started guide", "api reference", and "developer docs" where before it read "getting started guide", "api reference", "contributors and developers", "programming guide", and "tutorial: adding an executor". I think it makes a bit more sense to be organized in the new way but I don't think it's a big deal.

hategan (approved):
I'd only be wary of the website integration breaking on the two-layer docs. But if it doesn't, this looks mighty fine to me.

jameshcorbett:
It worked fine for me in my local build. I'll leave the PR up until next week though in case anyone else wants to comment.

PR #290, opened 2022-09-19
Fix firefox iframe scroll

hategan:
Firefox does not properly scroll to anchors in an iframe if the iframe is
very large. There appear to be similar issues in Firefox which can be fixed
by explicitly calling `scrollIntoView()` on the target element. This
solution appears to work here, too.

The problem manifests itself if you go to, e.g., the API ref for Cobalt and
try to either click on permalinks for the various methods or attempt to
navigate to the "See <base class method>" links.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,7 +34,7 @@ html_sidebars = {'**': ['globaltoc.html', 'relations.html', 'sourcelink.html', '
 
 # These are needed for the dhtml trickery
 html_static_path = ["_static"]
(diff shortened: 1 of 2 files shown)


hategan:
I checked in both Firefox and Chrome that clicking on doc links and method permalinks works as expected, but with browsers it's always a quagmire.

jameshcorbett (approved):
Looks reasonable to me, and I have run into the same problem. I'm not a JavaScript expert though.

andre-merzky (approved):
Thanks for addressing this - LGTM, but I am no JS expert either...

PR #291, opened 2022-09-20
Removed mentions of psij-ci-load since it has been phased out by

hategan:
psij-ci-env. Psij-ci-run still attempts to load it if it's there just in case.

### DIFF
--- a/psij-ci-run
+++ b/psij-ci-run
@@ -16,7 +16,10 @@ failifempty() {
     fi
 }
 
-source psij-ci-load
+if [ -f psij-ci-load ]; then
+    source psij-ci-load
+fi
+
 if [ -f psij-ci-env ]; then
     source psij-ci-env
 fi
--- a/psij-ci-run
+++ b/psij-ci-run
@@ -16,10 +16,6 @@ failifempty() {
     fi
 }
 
-if [ -f psij-ci-load ]; then
-    source psij-ci-load
-fi
-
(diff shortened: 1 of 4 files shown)


psij-ci-run
```diff
@@ -16,7 +16,10 @@ failifempty() {
     fi
 }
 
-source psij-ci-load
+if [ -f psij-ci-load ]; then
+    source psij-ci-load
+fi
```
jameshcorbett, line 21: I'm not completely sold on keeping the invocation of `psij-ci-load`. Any user who put anything in their psij-ci-load is going to get a conflict when they pull this commit (I think) and so should figure out that something has changed. How about putting a warning message instead? Or keeping what you have but also echoing a deprecation warning.

jameshcorbett (approved):
Thanks @hategan , LGTM. One minor comment below.

psij-ci-run
hategan: I'm thinking that any user who has a working setup with something in `psij-ci-load` can and should probably keep it as is.

That said, I'm not completely sold on keeping it either, but for a slightly different reason: if we need to instruct a user to do some fancy setup requiring something like `psij-ci-load`, we could as well instruct them to put the relevant lines in `psij-ci-run`.

hategan: So maybe there's consensus here that this should be removed.

PR #294, opened 2022-10-06
Repo rename prep

hategan:
*PLEASE DON'T MERGE AFTER REVIEW*

This PR is in preparation for renaming the repo to "psij-python". It is needed because existing tests deployments will continue to reference the old repo. This PR modifies the test runner to patch relevant files and update to the new URL automatically.

The plan, once this is reviewed, is as follows:
- merge this
- immediately:
    - rename the repo
    - clone the repo to also have a copy in the old location so that running tests can get the updated test runner
    - (alternatively, we could clone the old repo, but it may be better to do a rename to ensure that all ancillary data -- e.g. wiki pages -- is preserved)
- add a warning to the clone (old URL) that the repo is not used any more and that existing checkouts should have their origin URLs updated
- archive the cloned repo (old URL) to prevent accidental writes/diverging development

### DIFF
--- a/tests/ci_runner.py
+++ b/tests/ci_runner.py
@@ -50,12 +50,13 @@ def read_conf(fname: str) -> Dict[str, str]:
     return conf
 
 
-def run(*args: str, cwd: Optional[str] = None) -> None:
+def run(*args: str, cwd: Optional[str] = None) -> str:
     p = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, check=False,
                        cwd=cwd, text=True)
     if p.returncode != 0:
         print(p.stdout)
         raise subprocess.CalledProcessError(p.returncode, args, output=p.stdout)
+    return p.stdout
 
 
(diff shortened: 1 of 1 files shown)


tests/ci_runner.py
```diff
@@ -156,6 +157,44 @@ def run_tests(conf: Dict[str, str], site_ids: List[str], branches: List[str], cl
...
 
+OLD_REPO = 'ExaWorks/psi-j-python'
+NEW_REPO = 'ExaWorks/psij-python'
+
+
+def patch_file(file_name: str) -> None:
+    if os.path.exists(file_name + '.is_patched'):
+        return
+
+    with info('Pathcing %s' % file_name):
```
andre-merzky, line 168: small typo: `Pathcing` -> `Patching`

andre-merzky (approved):
The proposed approach for the rename sounds good to me.  I would specifically agree that renaming the repo to preserve issues, wiki etc. is much preferable to cloning.

PR #296, opened 2022-10-09
`git fetch remote` fetches the remote into FETCH_HEAD not into origin…

hategan:
…/main.

So the line that follows doesn't really do anything. This fixes it to do the checkout from the nearly fetched data. Yay git!


andre-merzky (approved):
Ha, I can imagine your thoughts when you found that one :-D

hategan:
> Ha, I can imagine your thoughts when you found that one :-D

Basically "I miss SVN".

PR #300, opened 2022-10-11
Updated name/references to `psij-python` from `psi-j-python`.

hategan:


andre-merzky (approved):
Your changes look good @hategan .  However, I pushed another commit on top of yours with some additional replacements, would be good if someone else could also review this PR.

hategan:
> That's dedication

Really isn't. It's "I'm text-searching for 'psi-j' and I don't want to have to do the work of distinguishing between instances where it does matter and where it doesn't".

PR #301, opened 2022-10-11
Fix styling of selectors when using plain doc builds

### LINKED ISSUE #298: Selector tabs render badly on ReadTheDocs
As the title says.  For an example, see [here](https://psi-j-python--288.org.readthedocs.build/en/288/user_guide.html).  I think it's fine to render radio boxes in case tabs are not available, but we probably should remove the line breaks?

hategan:
The styling for selector tabs is only in the web build files. This PR adds relevant css to style selector tabs in vanilla doc builds. Fixes #298


andre-merzky (approved):
Thanks for taking care of this!

hategan:
Thanks for noticing the problem in the first place.

PR #303, opened 2022-10-12
Fixed issue in patching config file to update repo. If the repo was o…

hategan:
…n the

last line and the file had no new line at the end, the repo would become `psij-pytho`.


andre-merzky (approved):
This looks correct to me.  Another approach might be to sanitize input with `line = line.rstrip('\n')` and then do the replace.

hategan:
> This looks correct to me. Another approach might be to sanitize input with `line = line.rstrip('\n')` and then do the replace.

Yes, that sounds much better.

PR #304, opened 2022-10-14
Fix parsing of boolean config options, such as minimal_uploads.

hategan:
`store_true` for argparse doesn't actually look at value, only at presence,so `--option=false` is the same as `--option` and `--option=true`.

`store` only stores strings, so tests like `if minimal` would succeed for `--minimal-uploads=false`.

### DIFF
--- a/tests/ci_runner.py
+++ b/tests/ci_runner.py
@@ -59,9 +59,12 @@ def run(*args: str, cwd: Optional[str] = None) -> str:
     return p.stdout
 
 
-def get_conf(conf: Dict[str, str], name: str) -> str:
+def get_conf(conf: Dict[str, str], name: str, default: Optional[str] = None) -> str:
     if name not in conf:
-        raise KeyError('Missing configuration option "%s"' % name)
+        if default is not None:
+            return default
+        else:
+            raise KeyError('Missing configuration option "%s"' % name)
     return conf[name]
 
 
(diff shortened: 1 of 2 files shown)


jameshcorbett (approved):
Looks reasonable to me

jameshcorbett:
Although if you set `'store_true'` you get an error if an argument is provided: 

```
$ python ../../../ExaWorks/argtest.py --minimal-uploads=false
usage: argtest.py [-h] [--minimal-uploads]
argtest.py: error: argument --minimal-uploads: ignored explicit argument 'false'
```

From: 

```python
import argparse 


parser = argparse.ArgumentParser()

parser.add_argument('--minimal-uploads', action='store_true', default=False,
                 help='Enables minimal uploads mode, which restricts the information that '
                      'is uploaded to the test aggregation server. ')

args = parser.parse_args()

```

hategan:
> Although if you set `'store_true'` you get an error if an argument is provided:

Yeah, except you wouldn't do that unless it's invoked by `ci_runner.py` and then the output gets lost in a log that you never look at unless something goes terribly wrong.

hategan:
> > Although if you set `'store_true'` you get an error if an argument is provided:
> 
> Yeah, except you wouldn't do that unless it's invoked by `ci_runner.py` and then the output gets lost in a log that you never look at unless something goes terribly wrong.

But then maybe the problem is in how pytest is invoked (aka., in ci_runner) rather than in conftest. So let's not merge this just yet.

hategan:
Allright, let's try again. This time with the flag passed properly by ci_runner.py

PR #305, opened 2022-10-20
Release prep stuff

hategan:
Some minor fixes and additions in preparation for release. Please see individual commits for details.


tag_and_release.sh
```diff
@@ -2,10 +2,10 @@
 set -euf -o pipefail
 
 create_tag () {
-    if [ -z "$1" ]
-      then
+
+    if [ -z ${1+x} ]; then
```
jameshcorbett, line 6: My bash is limited but isn't this check always false? Because if `$1` is unset then it gets set to "x" before the `-z` test.

Also where does `-u` you mention in the commit message come in?

jameshcorbett (approved):
Looks reasonable, one comment below.

tag_and_release.sh
hategan:
```
${parameter:+word}
              Use Alternate Value.  If parameter is null or unset, nothing is substituted, otherwise the expansion of word is substituted.
```

I think what you say would happen with `${1-x}`.

The `-u` is at the top of the script (`set -euf`). Basically, with the old code, if you invoke the script as `tag_and_release.sh create_tag` without a version parameter, it fails with "unbound parameter $1" rather than setting `VERSION` to "unknown" as seemingly intended. It's possible that the test was there to catch a situation where the script would be invoked as `tag_and_release.sh create_tag ""`, but I doubt it.

jameshcorbett: Ah OK, thanks for the explanation.

hategan: Thanks for asking. I actually had no idea until you asked. Found the solution on the googles, tested it, and didn't think twice. It's not clear to me from the documentation that it would work, since it hinges on "nothing" being the empty string in "nothing is substituted".

yadudoc (approved):
This looks good to go. @hategan let me know when you'd like me to push to pypi.

PR #310, opened 2022-10-23
Feature/rest service

andre-merzky:
This PR provides an exemplary REST service implementation for PSIJ, equivalent to the ZMQ service implementation in #293.  In fact, the executors and the service scripts have a fairly small code diff.

I open this PR for feedback and discussion again, not necessarily to merge soon.  I would actually propose *not* to merge either PR into PSIJ proper, but instead would like to suggest to use this opportunity to demonstrate how external executors can be contributed.  We should thus move the respective codes into separate repositories, IMHO.  That has the added advantage that we don't pollute the PSIJ dependencies with fastapi, websockets, zmq, radical.utils, etc.

 PS: I forgot to mention that this branch is based off `feature/zmq_service` instead of `main` - thus the PR target.
 
 PPS.: I should also mention that this is a rather crude implementation: no AA implemented, client IDs should be replaced with cookies, etc.  It's really more a prove of concept than anything else.  Noteworthy though is that notifications are implemented via websockets and seem to work as expected.


scripts/rest_service.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python3
+
+import uvicorn
+
+
+'''
```
jameshcorbett, line 6: Docstring should go before imports

scripts/rest_service.py
```diff
@@ -0,0 +1,300 @@
...
+
+        self._clients: Dict[str, _Client] = dict()
+        self._deserialize = psij.Import()
+        self._log = logging.getLogger('psij')
+        self._cnt: int = 0
+
+        # ----------------------------------------------------------------------
+        # websocket endpoint at which client can regoister for state updates
+        @app.websocket("/ws/{cid}")
+        async def ws_endpoint(ws: WebSocket, cid: str) -> None:
```
jameshcorbett, line 115: Is this async function used anywhere? I don't see it referenced anywhere. Is the decorator doing some magic?

scripts/rest_service.py
```diff
@@ -0,0 +1,300 @@
...
+    def _request_cancel(self, cid: str, jobid: str) -> None:
+        '''
+        parameters:
+           cid:str   : client ID
+           jobid:str : psij job ID for job to be canceled
+        '''
+
+        client = self._clients.get(cid)
+        if not client:
+            raise ValueError('unknown client cid %s' % cid)
```
jameshcorbett, line 212: Maybe just call ``self._clients[cid]`` and let a KeyError be thrown? Not important though

src/psij/executors/rest_service.py
```diff
@@ -0,0 +1,163 @@
...
+
+        :param url: address at which to contact the remote service.
+                    Supported schemas: `http://` and `https://`
+        :param config: The `RestServiceJobExecutor` does not have any
+                    configuration options.
+        """
+        if not config:
+            config = JobExecutorConfig()
+
+        if not url:
```
jameshcorbett, line 51: Maybe better to require a url?

src/psij/executors/rest_service.py
```diff
@@ -0,0 +1,163 @@
...
+        self._cid = str(requests.get('%s/executor/%s' % (url, name)).json())
+
+        # create a daemong thread for websocket state notifications
+        t = threading.Thread(target=self._state_listener)
+        t.daemon = True
+        t.start()
+
+    def __del__(self) -> None:
+        """Stop websocket thread upon destruction."""
+        ...
```
jameshcorbett, line 87: should be ``pass`` I think, or else nothing (it's not a syntax error as long as there's a docstring). Not that it's an error as is though

jameshcorbett: But since this method isn't used, should just be removed?

jameshcorbett (approved):
Looks really cool, thanks @andre-merzky !  Some silly comments below which you are free to ignore.

scripts/rest_service.py
andre-merzky: Indeed, the decorator does some magic - the function is handled in the `app` event loop.

scripts/rest_service.py
andre-merzky: I prefer the explicit error message in this case.

src/psij/executors/rest_service.py
andre-merzky: Yeah, that's fair - I'll remove the default.

src/psij/executors/rest_service.py
andre-merzky: Oh, right, that's a leftover from a different approach.  Thanks, removed it now.

scripts/rest_service.py
andre-merzky: Thanks - I moved the import.

scripts/rest_service.py
```diff
@@ -0,0 +1,303 @@
...
+        returns:
+            str: unique ID identifying the registered client
+        '''
+
+        # register new client
+        cid = 'client.%04d' % self._cnt
+        self._cnt += 1
+
+        # create executor
+        jex = psij.JobExecutor.get_instance(name=name, url=url)
```
hategan, line 175: This works for now. What we might want to consider is different executor configs. The other thing that might be relevant is that either `JobExecutor.get_instance` guarantees referential transparency, or we cache executor instances here.

scripts/rest_service.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # register state callback for this cid
+        cb = functools.partial(self._status_callback, cid)
+        jex.set_job_status_callback(cb)
+
+        # store client information
+        self._clients[cid] = _Client(jex)
```
hategan, line 182: I can't find a place where clients are removed from `_clients`. A client-facing call named `register` would suggest that a complementary `unregister` might be there. If that's the solution going forward, I'd be wary, since clients can disappear at any time and there are no guarantees that proper cleanup is done.

I would generally favor either cleaning based on job state (if a state if final, unregister, but that might pose some problems with polling) or finding a stateless solution.

hategan (commented):
Looks great. Some minor comments inside about general design stuff that I'm assuming are not supposed to be finalized anyway.

scripts/rest_service.py
andre-merzky: Ack

scripts/rest_service.py
jameshcorbett: Seems reasonable

scripts/rest_service.py
andre-merzky: Fair, an `unregister` would make sense, even if just for symmetry reasons, I'll add it (also to the ZMQ implementation where the same applies).  

As for cleaning idle clients after a while: we should then also allow to reconnect, e.g. support registering the same client ID again as a way to reconnect.  But it also might be sufficient for now to leave that problem to be resolved on the client side: if a call with an outdated client ID fails because the client was cleaned out after a while, the client should still be able to re-register for a new ID and then re-attach all it's job IDs.  This could be implemented in the client's REST executor.  That assumes that the backend executor can perform that reattach call, which I think is true for all executors we have (and is in fact the very purpose of that reattach anyway).  Would that work for you?

andre-merzky:
@jameshcorbett , @hategan : do you have any thoughts about this part:

>  I would actually propose *not* to merge either PR (ZMQ and REST) into PSIJ proper, but instead would like to suggest to use this opportunity to demonstrate how external executors can be contributed. We should thus move the respective codes into separate repositories, IMHO. That has the added advantage that we don't pollute the PSIJ dependencies with `fastapi`, `websockets`, `zmq`, `radical.utils`, etc.

Thanks :-)

scripts/rest_service.py
hategan: Yes, something along those lines. Would probably need some thinking and design work to smooth out.

jameshcorbett:
I think it sounds fine, but I think the simplicity of having everything in the same repo might outweigh the benefit of demonstrating how external executors can be contributed. Also, couldn't the additional third-party libraries be added in an extension?

hategan:
> @jameshcorbett , @hategan : do you have any thoughts about this part:
> 
> > I would actually propose _not_ to merge either PR (ZMQ and REST) into PSIJ proper, but instead would like to suggest to use this opportunity to demonstrate how external executors can be contributed. We should thus move the respective codes into separate repositories, IMHO. That has the added advantage that we don't pollute the PSIJ dependencies with `fastapi`, `websockets`, `zmq`, `radical.utils`, etc.
> 
> Thanks :-)

I think modularity is a great idea. And I also think that, since designing/implementing Layer 1 is a complex business, having some sandboxes to play with and learn and have others try without affecting the core psi/j distribution is likely a great way to go about it.

src/psij/executors/rest_service.py
jameshcorbett: Looks good although I think instead of checking if not url you could just make it a required/positional argument.

hategan:
> the simplicity of having everything in the same repo

Perhaps, but I think, in practice, it would be the difference between `pip install psij-python` and `pip install psij-python psij-rest-service`.

jameshcorbett:
I don't have strong feelings about it, I think two repos is totally fine.

hategan:
> I don't have strong feelings about it, I think two repos is totally fine.

I don't either. I have various opinions about various aspects, but I don't think they move us closer to one or the other.

andre-merzky:
Thanks both - then let me factor this out into separate repos, I feel a bit more comfortable with that.

andre-merzky:
This moved to [here](https://github.com/ExaWorks/psij-rest/pull/1)

PR #311, opened 2022-10-26
Adding release instructions

yadudoc:
This is one potential release process, that gives you some guarantees of clean history given the volume of activity. I suspect that this might need some discussion before we are agreed on this.

Once this goes in, we can discuss whether it makes sense to do further automation, say once a tag is minted of a regex pattern, github action will do the push to pypi. I'm unsure of how safe this is because the multi-site testing infrastructure is completely separate from github actions and might need some human intervention to confirm everything is green.


hategan (approved):
We should add a note at some point about adding a readthedocs version, but we can do that later

wilke (commented):
Looks good.

PR #313, opened 2022-10-26
Adds a test that re-attaches a job, but to another executor instance …

hategan:
…than

the one used to create the job in the first place.

### DIFF
--- a/tests/test_executor.py
+++ b/tests/test_executor.py
@@ -32,17 +32,19 @@ def test_simple_job_redirect(execparams: ExecutorTestParams) -> None:
 
 
 def test_attach(execparams: ExecutorTestParams) -> None:
-    job = Job(JobSpec(executable='/bin/sleep', arguments=['1']))
-    ex = _get_executor_instance(execparams, job)
-    ex.submit(job)
-    job.wait(target_states=[JobState.ACTIVE, JobState.COMPLETED])
-    native_id = job.native_id
+    job1 = Job(JobSpec(executable='/bin/sleep', arguments=['1']))
+    ex = _get_executor_instance(execparams, job1)
+    ex.submit(job1)
+    job1.wait(target_states=[JobState.ACTIVE, JobState.COMPLETED])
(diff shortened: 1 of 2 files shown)


andre-merzky (approved):
Nice...

tests/test_executor.py
```diff
@@ -45,6 +45,21 @@ def test_attach(execparams: ExecutorTestParams) -> None:
...
+    ex.submit(job)
+    job.wait(target_states=[JobState.ACTIVE, JobState.COMPLETED])
+    native_id = job.native_id
+
+    assert native_id is not None
+    job2 = Job()
+    ex2 = _get_executor_instance(execparams)
+    ex2.attach(job2, native_id)
+    status = job2.wait(timeout=_get_timeout(execparams))
+    assert_completed(job2, status)
```
jameshcorbett, line 60: Maybe check that the two jobs have the same exit status?

jameshcorbett (approved):
Looks great, thanks!

tests/test_executor.py
hategan: There is no mechanism to guarantee that the statuses of the two jobs are both updated atomically. Since each job executor polls the queue at roughly 30s, it can take up to 30s for the two jobs to sync up in status. We could wait for the first job to be updated with a wait(), but then the check would be mostly redundant.

jameshcorbett: Yeah I was thinking:

```python

status = job.wait(...)
assert_completed(job, status)
```

or really anything equivalent to `assert job.wait(...) == job2.wait(...)`. Just as a sanity check that the same underlying job doesn't get two different outcomes. I could imagine something like that happening if we wrote the job completion check logic in some really weird way.

Not a big deal at all though.

hategan: That can be added.

hategan: Done. Does that work?

jameshcorbett: Yup!

PR #315, opened 2022-10-27
add environment type check to `JobSpec`

andre-merzky:


hategan (approved):
This looks good. Thanks!

It does not seem to handle the case when I do something like:

```python
spec.environment = {}
spec.environment['key'] = value
```

For some reason, which may be influenced by some things I've seen in the past in other projects and maybe here (#48), I had this faint expectation that the validation happen before submission, likely in the executor, possibly combined with other basic checks, and possibly put in the executor ABC.

I'm not quite saying we should do that, but maybe think about it.

andre-merzky:
Duh, I did not think of that case!  Bugger...
Well, I pushed an iteration which fixes that - but honestly that's getting a bit too much code for such a little thing....  What's your opinion?

> that the validation happen before submission, likely in the executor, possibly combined with other basic checks, and possibly put in the executor ABC.

Yeah, I first considered that also, but then liked that the exception is close to the code which causes it (job spec *creation*, not submission).  I'd be fine though to go that route also, specifically also because it gives us a space to add more checks without messing the code up all over the place.  Let me know if you'd prefer that, and I'll give it a try...

hategan:
> specifically also because it gives us a space to add more checks without messing the code up all over the place

Right? Like a `_validate_job()` in the executor ABC. Or maybe even more, like `_pre_submit()` (not that exact name though) which both validates and does things like `job.executor = self`, which is another thing that bothers me (because an executor implementation that forgets to do that will be hard to troubleshoot).

andre-merzky:
> Right? Like a `_validate_job()` in the executor ABC. Or maybe even more, like `_pre_submit()` (not that exact name though) which both validates and does things like `job.executor = self`, which is another thing that bothers me (because an executor implementation that forgets to do that will be hard to troubleshoot).

I guess that's the right way to go about it, agreed.  Let me look into it - but probably won't manage to do so before end of next week.  Thanks for the feedback!

hategan:
> but probably won't manage to do so before end of next week.

I've been playing around a bit with this. I would have liked a nice solution, such as typeguard validating object properties, but I guess that's not how it works. That said, would it may be useful to add `assert typeguard.check_argument_types()` to `JobSpec.__init__` and perhaps the other spec sub-objects? I know it would break some things, such as us quietly accepting `str` instead of `Path` in certain places, but I'd rather either reject str or add it to the type annotations.

If you want, I can also clean up my experiments and commit to this branch to implement the before-submit-job-checks, unless you have that already.

andre-merzky:
> > but probably won't manage to do so before end of next week.
> 
> I've been playing around a bit with this. I would have liked a nice solution, such as typeguard validating object properties, but I guess that's not how it works. That said, would it may be useful to add `assert typeguard.check_argument_types()` to `JobSpec.__init__` and perhaps the other spec sub-objects? I know it would break some things, such as us quietly accepting `str` instead of `Path` in certain places, but I'd rather either reject str or add it to the type annotations.
> 
> If you want, I can also clean up my experiments and commit to this branch to implement the before-submit-job-checks, unless you have that already.

Hey @hategan - I am out of business this week I am afraid, so no updates from my end.  

The typecheck approach sounds appealing to me (although I have heard people argue that it's not pythonic :-P).  I would be a bit careful about runtime overhead (relevant when we at some point in the future start our phase-3), but that's the only word of caution from my side.  Thanks!

hategan:
> > If you want, I can also clean up my experiments and commit to this branch to implement the before-submit-job-checks, unless you have that already.
> 
> Hey @hategan - I am out of business this week I am afraid, so no updates from my end.

So I have something ready to go, but I don't want to step on your toes. We can wait for your changes, I can commit here, or do a separate PR.

andre-merzky:
> So I have something ready to go, but I don't want to step on your toes. We can wait for your changes, I can commit here, or do a separate PR.

You would not be stepping on my toes - you would be helping me out! :-)  So thanks, and feel free to push!

andre-merzky (commented):
Thanks @hategan , looks good from my perspective.  (Can't approve as I opened the PR)

hategan:
> Can't approve as I opened the PR

An unexpected plot twist! You can comment that you approve, and I will push the button for you. Or maybe @jameshcorbett can give it a quick look.

jameshcorbett:
Happy to do that.

jameshcorbett (approved):
LGTM. Although won't the new `check_argument_types` calls raise a lot more new errors than are covered by the new tests? Should we maybe add some test cases for them?

hategan:
> LGTM. Although won't the new `check_argument_types` calls raise a lot more new errors than are covered by the new tests? Should we maybe add some test cases for them?

Wouldn't those be tests for typeguard rather than psij code?

jameshcorbett:
> > LGTM. Although won't the new `check_argument_types` calls raise a lot more new errors than are covered by the new tests? Should we maybe add some test cases for them?
> 
> Wouldn't those be tests for typeguard rather than psij code?

Yes, sort of. I suppose they would be tests that we are using typeguard properly and that it is doing what we want it to. Mostly I was just thinking: this is new functionality/behavior, we should test it.

jameshcorbett:
But merge the PR, we can do it later.

PR #323, opened 2022-12-05
LLNL Tests for Getting Started Section of Documentation

RamonAra209:
These tests refer to the code snippets available at: https://exaworks.org/psij-python/#docs/getting_started.html


hategan (changes requested):
Thanks. This is a great start.

However, I am not sure that this will run when the main pytest suite is run.

Please take a look at the existing tests to see how they deal with having different executors. For example, a simple test in the suite is the following:

```python
from executor_test_params import ExecutorTestParams
from _test_tools import _get_executor_instance, _get_timeout, assert_completed, _make_test_dir


def test_simple_job(execparams: ExecutorTestParams) -> None:
    job = Job(JobSpec(executable='/bin/date', launcher=execparams.launcher))
    ex = _get_executor_instance(execparams, job)
    ex.submit(job)
    status = job.wait(timeout=_get_timeout(execparams))
    assert_completed(job, status)
```

This allows the test to run on all executors that are being tested. The test suite builds a list of executors that are available on the machine that the tests are run, and then iterates through all running each tests with each executor. The `execparams` pytest fixture takes care of that. 

I would suggest adding a very simple dummy test using the above fixture, running `make tests` to see what happens, and taking it from there. Also, feel free to ask if this is not clear enough.

RamonAra209:
> 
I see, I will reconfigure the tests accordingly, thank you.

hategan (approved):
Thanks Ramon! I don't see any problems with the code.

PR #325, opened 2022-12-24
Update WORKFLOW-EXAMPLES.md

kinow:


andre-merzky (approved):
Thanks for the fix!

andre-merzky:
@wilke : Hi Andres, any idea why the linter tests suddenly fail?  That is obviously unrelated to this PR - any suggestion on how to proceed?

hategan:
> any idea why the linter tests suddenly fail?

Looks like flake8 was updated to version 6.0.0, released on 11/23/22, which is now unable to tell that certain imports are used in type annotations (when only used in type annotations). Or something like that.

PR #326, opened 2023-01-07
Removed the creating your own fork part. It has not been followed by

hategan:
those who have write access to the repo and it is unnecessary for those. It is also implied for those who do not have write access to the repo (i.e., there is no other choice than to create a fork). So long story short, I don't think that part was necessary, enforced, or particularly useful.


andre-merzky (approved):
I agree with the rationale of removing this section.

PR #328, opened 2023-01-10
Downgrading flake8 to 5.x since 6.x messed up and is producing F401

hategan:
errors for things used in type annotations and docstrings.


andre-merzky (approved):
Thanks!

PR #329, opened 2023-01-11
Revert new tests

hategan:
Reverting this because it is freezing the CI on multiple machines.

One of the tests uses a shared global state without proper concurrency
controls, so it deadlocks on occasion, which leads to the test never
completing, which prevents the CI from running anything else. While the
CI should be resilient against such problems, the CI timing out a test
should be an issue of last resort.

This revert also removes a small number of other tests. The initial
idea that prompted this commit was to automatically test code that
is used in the documentation. However, this commit does not do so.
Instead, it adds some tests that are similar but not the same as what
is in the documentation, and are not automatically updated when
documentation tests are updated. Additionally, they do not test
anything the isn't already tested by the existing tests in the suite.


hategan:
> About the documentation tests: my memory may be confusing this with a discussion in a different project, but didn't we consider using sphinx's [`literalinclude`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) directive for documentation code so that it can be tested? Out of scope for this PR for this PR, just got triggered by the rationale...

That is also my recollection and I was under the impression that LLNL was working on that. I think it probably points to the fact that it may be wiser to mitigate risks by following the contributing guide and discussing most things in issues before too much work is invested. I filed a new issue (#330) so that we can stay on track with what was discussed initially.

PR #336, opened 2023-01-20
Fix some typos detected by codespell + provide codespell workflow to catch possible future typos

yarikoptic:


docs/README-selectors.md
```diff
@@ -21,7 +21,7 @@ substitute the value selected in the selector. For example:
     x = f("<&executor-type>")
 
 When rendered, the `<&executor-type>` string will be replaced by the
-lower case version of the selected value in the immediately preceeding
+lower case version of the selected value in the immediately prceeding
```
hategan, line 24: Should this be "preceding"?

docs/README-selectors.md
```diff
@@ -41,7 +41,7 @@ To add alternative code blocks, use:
 ```
 
 Alternate code blocks as well as value substitutions bind to the closest
-preceeding selector. However, selectors are synchronized, in that
+preeeding selector. However, selectors are synchronized, in that
```
hategan, line 44: ditto

hategan (commented):
Thank you, this is indeed useful.

docs/README-selectors.md
yarikoptic: Never trust a human - it was the one I had to fix manually. Sorry for the typo :-)
```suggestion
lower case version of the selected value in the immediately preceding
```

docs/README-selectors.md
yarikoptic:
```suggestion
preceding selector. However, selectors are synchronized, in that
```

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,17 @@
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
yarikoptic, line 8: I wonder though why this works did not run, may be some policy?

docs/README-selectors.md
hategan: Not sure I trust a computer much more :)

.github/workflows/codespell.yml
hategan: > I wonder though why this works did not run, may be some policy?

Yes. There's a button I see below saying "Approve and run [workflows]". Probably because they can run code.

hategan (approved):

PR #338, opened 2023-01-20
330 automatic test of some documentation code

RamonAra209:
Added one file per interactable snippet from the psij web documentation. Each file includes a pytestable version of its respective snippet.

These snippets were then rendered to the web-docs via using sphinx's literalinclude.


tests/getting_started/test_scheduling_information.py
```diff
@@ -0,0 +1,18 @@
...
+
+def test_getting_started_scheduling_info(execparams: ExecutorTestParams) -> None:
+    executor = JobExecutor.get_instance(execparams.executor)
+
+    job = Job(
+        JobSpec(
+            executable="/bin/date",
+            resources=ResourceSpecV1(node_count=1),
+            attributes=JobAttributes(
+                queue_name="<QUEUE_NAME>", project_name="<ALLOCATION>"
```
hategan, line 13: Maybe comment this line out until we get a way of passing these types of parameters from the test config file.

hategan (changes requested):
Beautiful. Just what was needed.

There are some type annotations missing and a line that will cause the tests to fail (see below, it can be commented out), but otherwise this is ready to go.

PR #339, opened 2023-01-27
Upgrade flake8 use typedecl

hategan:
flake8 6.0.0 (and corresponding pyflakes) removed support for `# type: abc`, which caused unused import errors. The solution was to change those type declarations to the "proper" format (`var: <type> = <value>`).


hategan:
I didn't branch this out of main, did I

PR #340, opened 2023-01-27
Upgrade flake8 use typedecl 2

hategan:
flake8 6.0.0 (and corresponding pyflakes) removed support for `# type: abc`, which caused unused import errors. The solution was to change those type declarations to the "proper" format (`var: <type> = <value>`).

This time on top of main.


src/psij/executors/local.py
```diff
@@ -23,11 +23,11 @@ class _ProcessEntry(ABC):
...
         self.executor = executor
-        self.exit_code = None  # type: Optional[int]
-        self.done_time = None  # type: Optional[float]
-        self.out = None  # type: Optional[str]
+        self.exit_code: Optional[int] = None
+        self.done_time: Optional[float] = None
+        self.out: Optional[str] = None
         self.kill_flag = False
-        self.process = None  # type: Optional[subprocess.Popen[bytes]]
+        self.process: Optional[subprocess.Popen[bytes]] = None
```
andre-merzky, line 30: That's an interesting one - haven't seen that documented as type.  Where did you find this?

andre-merzky (approved):
LGTM, thanks

src/psij/executors/local.py
hategan: https://docs.python.org/3/library/subprocess.html#subprocess.Popen

It turns out Popen is a class. You'd think it's something like Open returning a File or, you know, Verb applied to Noun produces Noun, but no.

andre-merzky: Ah, sorry, I wasn't clear.  Yes, `Popen` is a class - but `classname[bytes]` is not a syntax I have seen anywhere before.

I found one [google hit](https://github.com/python/typeshed/issues/4948) by now which indicates that the Popen class is marked as generic - wasn't aware of that...

hategan: I see. Well, if you don't put the type parameters, mypy will produce an error. I'm not entirely sure where I found what that parameter is (str/bytes based on whether you open the output streams in text or binary mode), since I can't find a clear statement of that in the official python documentation.

andre-merzky: thx, good to know.

PR #343, opened 2023-02-03
This updates the main PSI/J image to:

hategan:
- raise the connectors through drop-shadows similar to the rest of the components
- re-arranges the schedulers so that the arrows align nicely
- adds Gimp/Inkscape sources for the above

Old image:
psij

New image:
psij


andre-merzky (approved):
Oh thanks for fixing that!  Those misaligned arrows irked me, too - but other than you I was too lazy to do anything about it... :-/

PR #344, opened 2023-02-04
Check if the job is in a final state before attempting to ask the exe…

hategan:
…cutor

to cancel it.

### DIFF
--- a/src/psij/job.py
+++ b/src/psij/job.py
@@ -154,12 +154,10 @@ class Job(object):
 
         :raises psij.SubmitException: if the job has not yet been submitted.
         """
+        if self.status.final:
+            return
         if not self.executor:
             raise SubmitException('Cannot cancel job: not bound to an executor.')
-        elif self.status.final:
-            # TODO: we probably need a different type of exception for this and the above case,
-            # TODO: but this is probably a spec issue
-            raise SubmitException('Cannot cancel a job in a final state')
         else:
             self.executor.cancel(self)
 
--- a/src/psij/job.py
+++ b/src/psij/job.py
@@ -156,6 +156,10 @@ class Job(object):
         """
         if not self.executor:
             raise SubmitException('Cannot cancel job: not bound to an executor.')
+        elif self.status.final:
+            # TODO: we probably need a different type of exception for this and the above case,
+            # TODO: but this is probably a spec issue
+            raise SubmitException('Cannot cancel a job in a final state')
         else:
             self.executor.cancel(self)
 


andre-merzky (commented):
The spec is actually silent about when `cancel()` would raise a `SubmitException`.  But the definition of the exception type says: 

> `SubmitException`: Thrown if the request cannot be sent to the underlying implementation. [...]

As such I agree this is not the right exception to raise.

My question would be why we raise at all?  The `cancel()` semantics is defined as requesting job cancellation from the backend, but no guarantee of that cancellation occuring before the call returns.  I agree that that request is useless if the job is already final - but I think it might be simpler to just `pass` the request in that case instead of raising an exception.

with 'simpler' I mean the user perspective: with the proposed change one would need to `try/except` all `cancel` calls as there is always a race.

Current version:
```py
# only cancel if job is not final to avoid backend request on final jobs
if not job.status.final:
    # races here
    job.cancel()
    job.wait()
````

With the proposed change one would always need to write:
```py
try:
    job.cancel()
except psij.SubmitException:
    # only ignore the exception if job is already final
    if not job.status.final:
        # note: we just raced the state check again
        raise
job.wait()
```

I would much prefer this:
```py
# cancel job in case it is not final
job.cancel()
job.wait()
```

What do you think?

hategan:
> My question would be why we raise at all?

Good point, and what you say is rather convincing. There is one simple reason why I'd do this though. Something, down the line, will raise an exception, whether it's the local executor not finding the job in its array or `qcancel` exiting with some non-zero exit code, etc. However, in those cases, the exception tends to be rather obscure. And that last part is the main motivation. I was writing some tutorial stuff, and Jupyter printed this nested stack trace with lots of code, in pytest style, that ended in some rather obscure "no such key".

There is also some precedent to this, I think. We do some job sanity checks in `submit()` for the same reason: the error messages are consistent and clear.

andre-merzky:
> Something, down the line, will raise an exception, whether it's [this or that]

On this I certainly agree, but internal enactment errors and incorrect state errors are orthogonal issues, aren't they?  More precisely:

  1. An application is calling `job.cancel()` on a job, and the psij is able to register that cancellation request with the execution backend (synchronously or asynchronously).  All is well, the job may or may not be final when the call returns.
  2. An application is calling `job.cancel()` on a job, but the psij implementation encounters an internal error (job not found, backend `qcancel` error, etc).  If that happens synchroneously (*during* the cancel call), then yes, that should get translated into a non-obscure `psij.Eexception` meaning `the request to cancel failed` (right now `SubmitException`, but possibly something better at some point).

The fact if the job we call the `cancel()` on is final or non-final does not really play a role here, does it?  If the above holds, then

```py
class Job:
    def cancel(self):
        if self.status.final:
            return
       # do the deed
       ... 
```

is a convenient optimization which is sure to maintain the promised call semantics.

hategan:
> The fact if the job we call the `cancel()` on is final or non-final does not really play a role here, does it? If the above holds, then
> 
> ```python
> class Job:
>     def cancel(self):
>         if self.status.final:
>             return
>        # do the deed
>        ... 
> ```
> 
> is a convenient optimization which is sure to maintain the promised call semantics.

Now I understand. And agree.

andre-merzky (approved):
Thanks, LGTM

hategan:
Thank you!

PR #346, opened 2023-02-07
Check that the job is in a new state before submitting.

hategan:
Spent one hour on the tutorial trying to figure out why my job was completed before the output was generated. Well, it's because it was a previously completed job that was being submitted again, but wait() didn't know that.


andre-merzky (approved):
Ha, I can imagine how surprising that was :-P
LGTM.

PR #354, opened 2023-03-07
Added a note about relative/absolute paths for job files and some of …

### LINKED ISSUE #229: Add docs note about restrictions for paths for output files
When redirecting job output, one must generally specify an absolute path where the files should go (although #228 aims to change that).

However, this is not sufficient to guarantee that the output files will be produced at the specified location. Two more conditions need to be true:
- the path is on filesystem accessible on compute nodes
- the path must point to the same file on both submit node and compute node (i.e., the shared fs cannot be mounted under a different mountpoint on the submit vs. compute node)

Add notes to this extent in the API docs for job spec and other relevant guides.

### LINKED ISSUE #353: Document the subtleties associated with using relative/absolute paths
We had a bit of a discussion on #228 on this topic and the results should be in the documentation.

hategan:
…the issues involved in redirection.

Fixes #353.
Also fixes #229

### DIFF
--- a/src/psij/job_spec.py
+++ b/src/psij/job_spec.py
@@ -23,6 +23,45 @@ class JobSpec(object):
         """
         Constructs a `JobSpec` object while allowing its properties to be initialized.
 
+        .. note::
+            A note about paths.
+
+            It is strongly recommended that paths to `std*_path`, `directory`, etc. be specified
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
Thanks @hategan , that looks complete and informative to me.

PR #355, opened 2023-03-08
Not sure what happened there

### LINKED ISSUE #341: Missing ref in documentation for available launchers
This might be obsolete but the current API documentation does not point to a list of available launchers in class JobSpec.

launcher ([Optional](https://docs.python.org/3/library/typing.html#typing.Optional)[[str](https://docs.python.org/3/library/stdtypes.html#str)]) – The name of a launcher to use, such as “mpirun”, “srun”, “single”, etc. For a list of available launchers,:ref:launchers

hategan:
Fixes #341


andre-merzky (approved):
Thanks!

PR #361, opened 2023-03-15
Fix text squeezing on large displays.

hategan:
On large displays (with width >>14in), when viewing the documentation pages, the padding kicks in. It is meant to shift the text towards the center while preserving the width of the main text (through the max-width property). However, the padding is part of the width, so it ends up reducing the usable text width.

We could switch to controlling the margin, but then the thin left border also moves, which is unpleasant. I guess we could do that AND have the sidebar draw the border, but that's two steps.

So anyway, this adjusts the max-width to take the padding into account.

This addresses https://github.com/ExaWorks/exaworks.github.io/issues/5

PS: checks fail due to #362

### DIFF
--- a/web/docs/_static/cloud.css
+++ b/web/docs/_static/cloud.css
@@ -106,9 +106,9 @@ div.body {
     line-height: 1.5em;
     padding: 30px 20px;
     min-width: 0;
-    max-width: 8in;
+    max-width: calc(8in + max((100% - 14in) / 2, 8pt));
     border-left: 1px solid var(--color-fg-lines);
-    padding-left: max(calc((100% - 14in) / 2), 8pt);
+    padding-left: calc(max((100% - 14in) / 2, 8pt));
 }
 
 div.sphinxsidebar {


andre-merzky (approved):
I am not sure what the inversion of `max(calc(),...)` to `calc(max(...))` does, but trust that this is intentional.  Looks good to me.

mtitov:
May be set `max-width` as following `calc(max(8in, 100% - max((100% - 14in) / 2, 8pt)));`?

p.s. current update makes it look better than it is now, but there is still a lot of empty space on a right side, that's why in my proposal I've added `100% -`

hategan:
> I am not sure what the inversion of `max(calc(),...)` to `calc(max(...))` does, but trust that this is intentional. Looks good to me.

It's for consistency with max-width mostly and to essentially have all calculations under a single `calc()`

hategan:
> May be set `max-width` as following `calc(max(8in, 100% - max((100% - 14in) / 2, 8pt)));`?
> 
> p.s. current update makes it look better than it is now, but there is still a lot of empty space on a right side, that's why in my proposal I've added `100% -`

The limited width is intentional and is understood to help with legibility. See, e.g., https://practicaltypography.com/line-length.html. The current line length is a bit less than 90 characters, which is the upper limit that the fellow that wrote that document recommends. See https://baymard.com/blog/line-length-readability for another, similar, suggestion.

hategan:
> > The limited width is intentional and is understood to help with legibility. See, e.g., https://practicaltypography.com/line-length.html. The current line length is a bit less than 90 characters, which is the upper limit that the fellow that wrote that document recommends. See https://baymard.com/blog/line-length-readability for another, similar, suggestion.
> 
> oh, that the next level for me :) thank you for explanation and links!

On the other hand, I'm not sure wikipedia agrees: https://en.wikipedia.org/wiki/Line_length
That said, I'm not quite sure wikipedia agrees with wikipedia on that one.

PR #363, opened 2023-03-15
Pin typeguard version to 2.xx.

### LINKED ISSUE #362: Typeguard 3.0 removes check_argument_types
... which breaks PSI/J.

They say "use the typechecked decorator", but I believe we were using check_argument_types because we could have it in an assert line which could be disabled by users when python is invoked with `-O`.

hategan:
Also pins filelock and psutil to known working major versions.

Fixes #362

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-filelock
-psutil
+filelock==3
+psutil==5
 pystache>=0.6.0
-typeguard
+typeguard==2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-filelock>=3<4
-psutil>=5<6
+filelock~=3.4
+psutil~=5.9
 pystache>=0.6.0
-typeguard>=2<3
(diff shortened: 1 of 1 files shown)


hategan:
I'm merging this without review as an emergency measure because, without it, the code is mostly unusable and because actions, such as web site deployment, fail.

hategan:
A second set of eyes would still be useful here though.

andre-merzky (commented):
Ack on the pin for typeguard.  I also don't mind the version pin on filelock and psutil, but what is the rationale?  Do you consider pinned versions more reliable / reproducible / stable in general?

hategan:
> Ack on the pin for typeguard. I also don't mind the version pin on filelock and psutil, but what is the rationale? Do you consider pinned versions more reliable / reproducible / stable in general?

I'm trying to avoid the same problem happening in the future with filelock and psutil and I think "compatible version" (i.e. ~=) is a reasonable solution that still allows security/stability updates.

PR #365, opened 2023-03-17
Some release related updates

hategan:
Some updated release instructions as well as the docs version using the psij version rather than a manual string.

This PR is a post-release update (i.e., the 0.9 release is done and it does not depend on this PR). However, the online docs still show "0.1.0-post1" as the version until this PR is merged.


andre-merzky (approved):
LGTM, thanks

PR #366, opened 2023-03-20
Install dependencies in the cloned dir rather than the dir where the …

hategan:
…tests

are set up from.

Also, make sure that dependencies are cleaned between branch tests since there are cases when pip doesn't downgrade correctly, which can lead to tests failing when they shouldn't.


andre-merzky (approved):
Makes sense to me.

PR #371, opened 2023-03-28
Fix 404 page (missing tabs property was making it render funny).

hategan:
Also added code to detect if the 404 is on a documentation page. The version switcher in the documentation attempts to preserve the path (so that if you are looking at, say, JobExecutor in a version, you'd also potentially be looking at JobExecutor after the switch). However, there is no guarantee that the docs structure is the same in the two versions, so you may land on a page that's not there. The code looks at whether you are hitting a 404 on a docs page and offers a link to the main docs page for that version.

Also, a minor fix in `getCurrentVersion`

### DIFF
--- a/web/404.html
+++ b/web/404.html
@@ -1,11 +1,20 @@
 ---
 permalink: /404.html
 layout: default
+tab: 100
 ---
 
 <div class="container">
-  <h1>404</h1>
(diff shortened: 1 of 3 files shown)


andre-merzky (approved):
Thanks

PR #373, opened 2023-04-11
Patch testing url

hategan:
This changes the test upload url in testing.conf to testing.psij.io (which is now live).

It also updates ci_runner.py to patch testing.conf on systems where the tests are already set up.

And ci_runner.py now has a slightly nicer way of applying such patches.

### DIFF
--- a/testing.conf
+++ b/testing.conf
@@ -130,7 +130,7 @@ custom_attributes =
 #
 # server_url = <url>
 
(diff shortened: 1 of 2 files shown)


hategan:
> I just wonder if using the `patch` utility would make this simpler?

Possibly, and we could do that. But probably not that much simpler.

PR #380, opened 2023-05-01
378 test queue parameter

RamonAra209:


hategan:
Thank you @RamonAra209!
There are some type and linting errors that should be fixed.

tests/test_queue.py
```diff
@@ -0,0 +1,129 @@
...
+    queues.extend(slurm_queues)
+    queues.extend(lsf_queues)
+
+    if len(slurm_queues) != 0:
+        scheduler = "slurm"
+    elif len(lsf_queues) != 0:
+        scheduler = "lsf"
+
+    if len(queues) < 2:
+        pytest.raises(Exception("Need at least two queues to perform this test"))
```
hategan, line 101: I think you are looking for `pytest.skip` instead.
`pytest.raises` is used to test that a block of code throws a specific exception. For example,

```python
def test_that_division_by_zero_correctly_raises_exception():
    with pytest.raises(ZeroDivisionError):
        1 / 0
```

In other words you use it to check that you test throws an exception. If you had `1 / 1` instead of `1 / 0`, `pytest.raises` would actually cause the test to fail because it did not throw the exception that was expected.

tests/test_queue.py
```diff
@@ -0,0 +1,129 @@
...
+
+    print("available queues:", queues)
+    test_queues = random.sample(queues, 2)
+    print("test queues:", test_queues)
+
+    executor = JobExecutor.get_instance(scheduler)
+
+    job1 = make_job(test_queues[0])
+    executor.submit(job1)
+    qstat = get_queue_info(scheduler)
```
hategan, line 112: Most of these commands accept a job id as an argument to only return info about a specific job.

tests/test_queue.py
```diff
@@ -0,0 +1,129 @@
...
+    job1_qstat_entry = [l for l in qstat if job1._native_id in l][0]
+    assert test_queues[0] in job1_qstat_entry
```
hategan, line 114: We might want to go a bit further than just checking if the queue name is somewhere in the qstat output. It's quite possible that we might have a queue named "test" and the word "test" appearing in an unrelated place in the qstat output.

hategan (changes requested):

tests/test_queue.py
```diff
@@ -0,0 +1,129 @@
...
+def get_slurm_queues() -> List[str]:
+    out = os.popen("mdiag -c").read().split("\n")
+    return [line.split("=")[-1] for line in out if "PartitionName" in line]
+
+
+def get_lsf_queues() -> List[str]:
+    valid_queues = []
+    out = "".join(os.popen("bqueues -u $(whoami) -o 'QUEUE_NAME NJOBS PEND RUN SUSP STATUS'").read()).split("\n")
+    out = [l for l in out if len(l) != 0]
+    out = [l.split(" ") for l in out]
```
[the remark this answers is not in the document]
RamonAra209: Done

tests/test_queue.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def get_slurm_queues() -> List[str]:
+    out = os.popen("mdiag -c").read().split("\n")
+    return [line.split("=")[-1] for line in out if "PartitionName" in line]
+
+
+def get_lsf_queues() -> List[str]:
+    valid_queues = []
+    out = "".join(os.popen("bqueues -u $(whoami) -o 'QUEUE_NAME NJOBS PEND RUN SUSP STATUS'").read()).split("\n")
```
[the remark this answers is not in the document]
RamonAra209: Done, now using the dictionary as intended

tests/test_queue.py
```diff
@@ -0,0 +1,129 @@
...
+    slurm_queues = get_slurm_queues()
+    lsf_queues = get_lsf_queues()
+
+    queues.extend(slurm_queues)
+    queues.extend(lsf_queues)
+
+    if len(slurm_queues) != 0:
+        scheduler = "slurm"
+    elif len(lsf_queues) != 0:
+        scheduler = "lsf"
```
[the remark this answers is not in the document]
RamonAra209: Can you take a look at my implementation? I kept the way I was detecting it, but am now only running the test when `execparams.launcher == "single"`

tests/test_queue.py
RamonAra209: Done, updated the `qstat` equivalents to query just for the job.

tests/test_queue.py
RamonAra209: Done, I applied some custom formatting on the `qstat` equivalents, so they now only report the queue name. This allows me to keep the existing assert logic

Ex: 
```
$ bjobs -o "queue" 4775749
QUEUE
pbatch
```

tests/test_queue.py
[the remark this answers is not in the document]
hategan: It's not about detecting the LRM on the system but the fact that we test multiple executors on that system. So even if you restrict it to the single launcher, it will still be repeated for the local, batch-test, and whatever PSI/J detected to be the scheduler.

You could remove execparams, but then you risk not having access to other necessary parameters that might be set by the users that set up the tests. By the way, you may want to use execparams.custom_attributes, since some systems require setting various things, like an account or project.

PR #381, opened 2023-05-10
Catch SIGCHLD to figure out when a child process completes instead of

hategan:
polling at fixed intervals of 0.2s. Without it, `/bin/sleep 0` would take a minimum of 0.2s to complete, which is unnecessary. With the change, the same takes 0.013s instead. Still too much if you ask me.

### DIFF
--- a/src/psij/executors/local.py
+++ b/src/psij/executors/local.py
@@ -147,7 +147,20 @@ class _ProcessReaper(threading.Thread):
 
     def _handle_sigchld(self) -> None:
         with self._cvar:
-            self._cvar.notify_all()
+            try:
+                self._cvar.notify_all()
+            except RuntimeError as ex:
+                # In what looks like rare cases, notify_all(), seemingly when combined with
+                # signal handling, raises `RuntimeError: release unlocked lock`.
+                # There appears to be an unresolved Python bug about this:
+                #    https://bugs.python.org/issue34486
+                # We catch the exception here and log it. It is hard to tell if that will not lead
+                # to further issues. It would seem like it shouldn't: after all, all we're doing is
+                # making sure we don't sleep too much, but, even if we do, the consequence is a
+                # small delay in processing a completed job. However, since this exception seems
+                # to be a logical impossibility when looking at the code in threading.Condition,
+                # there is really no telling what else could go wrong.
+                logger.warning('Exception in Condition.notify_all()', ex)
 
     def _check_processes(self) -> None:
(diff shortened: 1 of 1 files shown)


hategan:
Well, bummer. I'm getting a `RuntimeError: release unlocked lock` from `threading.Condition`:

```python
for waiter in waiters_to_notify:
           waiter.release()
```

This makes no sense since it indicates that there is a waiting thread that somehow isn't holding the cvar lock, except that `cvar.wait()` is inside a `with cvar:`.

I found this, https://bugs.python.org/issue34486, which is a 5 year old open issue, but suggests an issue with signal handling and locking in python.

So this is becoming a draft until we sort this out.

andre-merzky:
Ah, it seems you found the bottomless pit of python signal handling...

andre-merzky:
Ah, it seems you found the bottomless pit of python signal handling...   

I basically gave up on signalling in Python, back then when I looked at it it was broken all over the place with open tickets reaching back to 2008.  I just checked that ticket list I curated back then: some of the tickets are marked `resolved` (without further explanation though), but many still remain open, and I thus would consider signaling still broken in Python as long as *anything* from the `threading` module is used in the same code.

Ok, maybe I should try to be constructive...  The only way I can see this working would be to create a new child *process* which does not do any threading but which has wait's on a queue for spawn requests, spawns those processes, and waits on the queue again.  It also would have a signal handler which forwards any signal events on a different queue.  That way the parent process can outsource all process spawning and signal handling to that child and would get notifications about any cought `SIGCHLD` events via messages on the second queue.

But that is obviously adding complexity, makes the process tree somewhat non-intuitive, makes it harder to recover from failure etc etc. :-/

hategan:
It's annoying because I can't reproduce it on my laptop.

andre-merzky:
> It's annoying because I can't reproduce it on my laptop.

Is that possibly a newer Python version - maybe some fixes *did* make it into the Python releases at some point?

hategan:
> > It's annoying because I can't reproduce it on my laptop.
> 
> Is that possibly a newer Python version - maybe some fixes _did_ make it into the Python releases at some point?

3.6, so probably not. I can't even reproduce it any more in the github workflows.

hategan:
I added a try/catch around the issue since, even when that exception occurs, it things are seemingly no worse than without this PR (although read caveat in the comment around that catch block).

So maybe we should give this a try.

andre-merzky (approved):
LGTM!

PR #383, opened 2023-05-12
Don't hold the reaper lock while checking processes, since it can

hategan:
stall new jobs from being submitted. Instead, make a copy of the jobs dict and operate on that without holding the lock.

This seems to make the local executor faster (!) than plain Popen + wait, which I can't say I understand clearly.


andre-merzky (approved):
> which I can't say I understand clearly.

yeah, that's a bit surprising.  But numbers don't lie, so... `¯\_(ツ)_/¯`

PR #384, opened 2023-05-15
A preliminary implementation for the list() method for batch schedule…

hategan:
…r type

executors. The test should probably attempt to detect if headers are incorrectly being parsed as normal data lines, but it doesn't.


andre-merzky (approved):
Did not test-run, but the changes look very reasonable to me.  Thanks @hategan !

PR #385, opened 2023-05-23
Fix `Install from pip` instruction in getting started page of documentation

adi611:
The current `Install from pip` instruction in the Getting Started page of the documentation is:
```
pip install psij
```
which should instead be:
```
pip install psij-python
```


hategan:
Good catch! Thank you!

hategan (approved):

PR #388, opened 2023-05-24
Attach signal handler when starting reaper to avoid starting reaper on

hategan:
import. It also solves some problems with starting jobs from subprocesses.

However, a better solution for making things work properly in both parent process and subprocesses when fork is involved is to associate singleton threads with process IDs.


andre-merzky (approved):
good catch!

PR #389, opened 2023-07-18
Added a tutorial which was used before in various places.

hategan:


hategan:
To be clear, this is simply the PSI/J tutorial from https://github.com/ExaWorks/SDK

andre-merzky (approved):
Thanks for taking care of this!   It seems though that the notebook can't resolve the image path on top (`./images/psij_overview.png`)?

hategan:
I'll merge this for now. We should probably have a nice binder way of running this from out site. When that happens, we should probably make sure it's all nicely done.

Issue #387, opened 2023-05-24
Repeated jobs hang indefinitely when using multiprocessing with "fork" start method


tlhr:
I'm running into odd behavior when launching jobs from a separate process. The first execution seems to work fine, but an additional run in the same python kernel will cause a call to `job.wait()` to hang indefinitely. From a first glance it seems that the `_ProcessReaper.run()` method is not correctly running for subsequent jobs, causing the `job._status_cv` condition to never be notified for additional jobs.

Here is a short script that reproduces the problem:
```python
#!/usr/bin/env python3

import logging
from multiprocessing import Process, set_start_method
import psij


def func():
    # Get logs from PSI/J
    logger = logging.getLogger()
    logger.setLevel("DEBUG")
    logger.addHandler(logging.StreamHandler())

    exe = psij.JobExecutor.get_instance("local")
    job = psij.Job(psij.JobSpec("test", "echo", arguments=["foo"]))
    exe.submit(job)
    print(job, flush=True)

    # This hangs on the second round
    job.wait()
    print(job, flush=True)


if __name__ == "__main__":
    # `fork` hangs, `spawn` and `forkserver` work as expected
    set_start_method("fork")
    p = Process(target=func)
    p.start()
    p.join()

    print("Done, and again...")

    p = Process(target=func)
    p.start()
    p.join()

```

hategan:
Thanks for catching this.

The first process hangs for me.

hategan:
So the first process hangs for me because the reaper thread is started in the parent process but is not really running in the child processes. However, since forking involves a memory copy of the parent process, the variable that is meant to track that the reaper thread is started is set.

When I remove a key piece that lets the reaper be initialized in the subprocess, things seem to work.

So can you give me some more details about which version of PSI/J you are running and perhaps the system on which you are running this?

hategan:
PR #388 is meant to fix the first_process_hanging problem, but I don't think it's a solution for this issue.

tlhr:
> So can you give me some more details about which version of PSI/J you are running and perhaps the system on which you are running this?

Thank you so much for looking into this! The PSI/J version is pulled from master today, but it was also present in the pip package. The python version is 3.10, I can give you more system details tomorrow, I believe it's a recent Ubuntu LTS release.

hategan:
Can you give PR #388 a shot then?

It won't work with fork() if you submit from both parent and child, but it should work if you submit from child processes.

tlhr:
[...]

PR #390, opened 2023-07-21
Proper singleton threads

hategan:
This PR attempts to more gracefully deal with situations such as those in issue #387.

It does so by considering singleton threads to be associated with the current PID rather than a flag in memory, since values are
copied by `fork()`.

This is not really a solution for #387. The solution is, as advised by the `fork()` man page, to not use `fork()` with threads. It does not work and it is not meant to work.

The PR also moves the signal attachment back to the global scope of the local executor. The issue, as pointed out in #387, is that signal handlers in Python can only be attached from the main thread. Unfortunately, there is no way to guarantee, in Python, that a particular piece of code will only be called from the main thread. However, we can somewhat constrain the problem by requiring that psij is only imported from the main thread.

### DIFF
--- a/src/psij/executors/local.py
+++ b/src/psij/executors/local.py
@@ -18,8 +18,15 @@ from psij import JobExecutor
 logger = logging.getLogger(__name__)
 
 
+if threading.current_thread() != threading.main_thread():
+    raise ImportError('The psij module must be imported from the main thread.')
+
+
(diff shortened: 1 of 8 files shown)


andre-merzky (approved):
I like the approach and agree with the comments.  Yes, threading + multiprocessing in Python is an ugly mess and there seems to be no way to reconcile them completely.

PR #391, opened 2023-07-28
Added a simple check in `JobExecutor.submit()` to test if a job's

hategan:
executor property is already set (which would indicate that it was submitted before).


andre-merzky (approved):
Yeah, that should not happen indeed.

PR #392, opened 2023-07-28
Properly synchronize before incrementing current_job_index, otherwise

hategan:
we may end up skipping some jobs and submitting some twice.


andre-merzky (approved):
Makes sense, thanks.

Quick question on a second glance: do we really need an `RLock`?  I would think that a plain `Lock` would suffice?

PR #394, opened 2023-07-28
Fixed broken testing site image on web site.

hategan:


andre-merzky (approved):
Thanks for fixing this!

PR #395, opened 2023-08-02
Allow for `str` paths as inputs to `JobSpec`

arosen93:
Currently, the various path kwargs in `JobSpec`, such as `directory` and `stdout_path`, require the user to input them as `Path` objects. It would arguably be more convenient to allow the user to pass in any `PathLike` object, particularly a `str` representation. This PR simply calls `Path()` on these kwargs to automatically convert them to `Path` objects if they aren't already.

### DIFF
--- a/src/psij/job_spec.py
+++ b/src/psij/job_spec.py
@@ -1,12 +1,13 @@
+from __future__ import annotations
+
 import sys
 from pathlib import Path
-from typing import Optional, List, Dict, Any
+from typing import Any, Dict, List, Optional
 
 from typeguard import check_argument_types
 
 from psij.job_attributes import JobAttributes
 from psij.resource_spec import ResourceSpec
-
 from psij.utils import path_object_to_full_path as o2p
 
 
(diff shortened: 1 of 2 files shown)


andre-merzky:
Thanks @arosen93 !  I like the improved usability - but we would need to sync the spec for that, and that will need some discussion in the group and a bit of time, and we should also add a unit test.  Please allow us a week or so for getting back on this PR.

arosen93:
@andre-merzky makes sense! Keep me posted. Of course, I'd be happy to add a unit test if this PR ends up being desirable.

hategan:
Thanks @arosen93!
This has been the subject of an earlier discussion. There is indeed a gain in usability. Some of the downsides we've previously identified are that, besides updating the spec (which may be fairly easy since the spec is lenient as far as paths are concerned), we would need to also change the declared types of paths to `Path|str` as well as ensuring that the properties, if accessed directly, behave the same as the constructor keyword arguments.

That said, this does seem like a reasonable time to make all these changes and I would echo what @andre-merzky said earlier: let's take a bit of time to discuss, here, on the channel, and in meetings, the possible implications and go from there.

arosen93:
@hategan: Thanks for the reply! That certainly makes sense.

I have updated the type hints and added unit tests (although the full CI suite doesn't seem to run on PR, so I'll have to double-check locally in a little bit).  Regarding "updating the spec", I assume you are referring to some documentation that complements the `JobSpec` class? If so, let me know where I should be updating that, if at all. Anyway, this of course isn't anything urgent, but I do think it'd be a small convenience for folks.

hategan:
> @hategan: Thanks for the reply! That certainly makes sense.
> 
> I have updated the type hints and added unit tests (although the full CI suite doesn't seem to run on PR, so I'll have to double-check locally in a little bit).

It won't run for PRs from forked repos for security reasons. I approved it. You can also run it locally with `make tests`.

> Regarding "updating the spec", I assume you are referring to some documentation that complements the `JobSpec` class? If so, let me know where I should be updating that, if at all. Anyway, this of course isn't anything urgent, but I do think it'd be a small convenience for folks.

There's a formal specification for PSI/J at: https://github.com/ExaWorks/job-api-spec (and rendered at https://exaworks.org/job-api-spec/specification). I don't think it needs updating for this particular issue.

> 
> I think you were alluding to this above, but if a user does `JobSpec.directory = '/my/path'` it won't be a `Path` object, so that still needs to be addressed. Any recommendations for how you would like to proceed on that?

I think setters/getters should do the trick.

Again, thanks for the PR and all the feedback!

arosen93:
Fantastic, thanks! Tomorrow, I'll fix this up and add the setters/getters 😄

andre-merzky:
@hategan
> ensuring that the properties, if accessed directly, behave the same as the constructor keyword arguments.

I think you mean the setters, which makes sense.  But just to be on the safe site I wanted to state my opinion that the getters should always return `Path` type values.

hategan:
> @hategan
> 
> > ensuring that the properties, if accessed directly, behave the same as the constructor keyword arguments.
> 
> I think you mean the setters, which makes sense. But just to be on the safe site I wanted to state my opinion that the getters should always return `Path` type values.

Agreed. Sorry for not stating that explicitly.

hategan:
I was going to merge this, but I see that @andre-merzky may have had some concerns about the specification and the return type of the getters.

I believe that things are OK from both perspectives. The specification uses a `Path` class, but clarifies that implementations have some liberty in deciding what that class should be and I believe that convenience conversions like this are not necessarily a violation of the specification, especially in a language that does this kind of stuff routinely in its standard library. But I want to let @andre-merzky weigh in. As for getters returning `Path` rather than `str | Path`, that is what the updated code does.

andre-merzky (approved):
Yes, this is how it should be.  Thanks!

PR #403, opened 2023-08-10
Don't clean up submit file if it doesn't exist, which is the case when

hategan:
attaching a job to a native job.


andre-merzky (approved):
Makes sense of course.  thanks!

PR #406, opened 2023-08-22
Fix serialization

hategan:
This PR moves serialization outside of the core classes and implements it using introspection rather than manually creating dictionaries out of PSI/J objects.
The updated implementation also guarantees that `x == deserialize(serialize(x))`.


andre-merzky (commented):
The code looks correct to me.  

Question: there are a gazillion of Python serialization modules - why do we roll our own?

hategan:
> The code looks correct to me.
> 
> Question: there are a gazillion of Python serialization modules - why do we roll our own?

Can you be more specific?

We want something that serializes to JSON, for REST usage purposes as well as for non-python projects to easily interface with PSI/J-Python, and the Python JSON class does not serialize custom classes by default. This PR implements a general "convert to dict/list" feature which then allows easy conversion to various wire formats (e.g., JSON, XML, etc.). Of course, one can also implement serializers/deserializers that conform to the common interface in this PR by directly using some of the binary serialization libraries in Python (e.g., pickle) and bypassing the dict/list conversion.

andre-merzky (approved):
Thanks a bunch, looks great!  And ignore the question about serialization, got it now...

PR #408, opened 2023-08-29
Use auth keys for test uploads

hategan:
This is to go with https://github.com/ExaWorks/psij-testing-service/pull/57


andre-merzky (approved):
LGTM - thanks!

PR #410, opened 2023-09-02
Moved `name` parameter after `directory` so we can say `JobSpec('/bin…

hategan:
…/sleep', ['1'])`


andre-merzky (approved):
That makes sense!

PR #411, opened 2023-09-11
Docs improvements

hategan:
This is a rather large batch of changes that is centered around documentation improvements. I would recommend looking at individual commits, but I'll provide a summary here:
- quite a few updates to the structure and wording on the user guide, including moving all code examples to tests and using `.. literalinclude::` to insert them into the documents
- a remake of the API guide that combines the benefits of James' manual structure with automatic generation; this is done using
a skeleton document that gets expanded by a custom script
- web docs can now be made locally (`make web-docs-dev`)
- versioned web docs now include the dev version, but it is not selected for display by default unless it's the only one available (as would be the case with web-docs-dev).
- some minor-ish styling fixes for the web docs
- addition of the `queue_name` and `project_name` entries to `testing.conf` and their propagation through the pipes all the way to rendering them on the docs pages
- fixes to executor plugins allowing aliases (e.g. pbs and psbpro) as well as human-consumable labels on them (e.g. "Radical Pilot").
- fixed all references that sphinx complained about and the many other docstring fixes


mturilli (approved):
LGTM!

PR #412, opened 2023-09-12
Properly implement wait according to spec (i.e., wait for all states

hategan:
greater than `target_states` not just for `target_states`).

This goes a bit further by also implementing the yet-to-be-merged updates from https://github.com/ExaWorks/job-api-spec/pull/178


andre-merzky (approved):
Great - thanks for your patience with that discussion!

Issue #200, opened 2022-05-05
Nodelist


kylechard:
Libensemble would like to have access to a nodelist on different systems. They also have some code for several schedulers that could be used as a basis for this.

jameshcorbett:
See also [this issue in the job-api-spec repo](https://github.com/ExaWorks/job-api-spec/issues/148#issue-1216437288).

jameshcorbett:
In my tests on LLNL:

- Flux does not export any nodelist environment variables but there is a command you can run to get something like ``ruby[1-2]``
- Slurm gives you something like ``SLURM_JOB_NODELIST=ruby[1-2]`` which is fine
- LSF gives a bunch of stuff in different formats, but none of them is perfect:

```
LLNL_COMPUTE_NODES=lassen[23,26]
LSB_HOSTS=lassen710 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen26 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23 lassen23
LCSCHEDCLUSTER=lassen
LSB_SUB_HOST=lassen709
LSB_MCPU_HOSTS=lassen710 1 lassen26 40 lassen23 40
```

LLNL_COMPUTE_NODES is the best but not portable.

hategan:
So I posted this on Slack, but I'm adding here for posterity. It's from https://slurm.schedmd.com/srun.html.

`srun -l /bin/hostname | sort -n | awk '{print $2}' > $MACHINEFILE`

[...]

PR #414, opened 2023-09-14
Hostlist

hategan:
This makes (most) executors set a PSIJ_NODEFILE variable pointing to a nodefile, which is a solution for #200 and touches on https://github.com/ExaWorks/job-api-spec/issues/148, although it would be good to figure out a way to formalize this in the spec.

It also contains a few other minor fixes.


jameshcorbett (approved):
Looks reasonable to me! The format of all the node files is the same--a hostname on each line or something?

hategan:
> Looks reasonable to me! The format of all the node files is the same--a hostname on each line or something?

In theory, yes.

andre-merzky (approved):
I like that one :-)

PR #419, opened 2023-09-20
Test submit script generation basics

### LINKED ISSUE #416: `main` branch breaks `custom_attributes` (at least for Slurm)
On Python 3.10 on Linux, I am able to use the `custom_attributes` kwarg in `JobAttributes` as-expected on PSI/J 0.9.0 (from PyPI) but not on 3109e7c09d8ef6f46bbeac06890b101662d356c6. The only thing I am changing, to the best of my knowledge, is the PSI/J version on the remote machine.

The script I am running is as follows:

```python
from psij import Job, JobAttributes, JobExecutor, JobSpec, ResourceSpecV1

job_executor = JobExecutor.get_instance("slurm")
job = Job(
    JobSpec(
        name="test",
        executable="/bin/date",
        resources=ResourceSpecV1(node_count = 1),
        attributes=JobAttributes(project_name = 'matgen', custom_attributes = {'slurm.constraint': 'cpu', 'slurm.qos': 'debug'}),
        launcher="single",
    )
)

job_executor.submit(job)
native_id = job.native_id
print(native_id)
```

On PSI/J 0.9.0, everything works as expected. However, on the `main` branch, the `custom_attributes` do not get set. A comparison of the Slurm submit scripts generated by PSI/J are shown below.

0.9.0:

```bash
!/bin/bash



#SBATCH --export=ALL

#SBATCH --exclusive

#SBATCH --nodes=1





#SBATCH --time=0:10:00


#SBATCH --account="matgen"


#SBATCH --constraint="cpu"
#SBATCH --qos="debug"


#SBATCH -e /dev/null
#SBATCH -o /dev/null


exec &>> "/global/homes/r/rosen/.psij/work/slurm/$SLURM_JOB_ID.out"

/bin/bash /global/homes/r/rosen/.psij/work/f65b111560da52a6d41f52a732be0faf4914eb6d7af2230a8de326017979e684/single_launch.sh 4a4e3076-17f5-41da-97c0-b9d9bd8584e9 '' '' '' /dev/null /dev/null /dev/null /bin/date

echo "$?" > "/global/homes/r/rosen/.psij/work/slurm/$SLURM_JOB_ID.ec"
```

main:

```bash
#!/bin/bash



#SBATCH --export=ALL

#SBATCH --exclusive

#SBATCH --nodes=1





#SBATCH --time=0:10:0


#SBATCH --account="matgen"




#SBATCH -e /dev/null
#SBATCH -o /dev/null


exec &>> "/global/homes/r/rosen/.psij/work/slurm/$SLURM_JOB_ID.out"

/bin/bash /global/homes/r/rosen/.psij/work/f65b111560da52a6d41f52a732be0faf4914eb6d7af2230a8de326017979e684/single_launch.sh 772de4e0-5634-45ad-9ba9-920fa1f5643f '' '' '' /dev/null /dev/null /dev/null /bin/date

echo "$?" > "/global/homes/r/rosen/.psij/work/slurm/$SLURM_JOB_ID.ec"
```

In the latter case, I (unsurprisingly) get the following instead of the printed native ID:

```bash

Cell In[1], line 14
      3 job_executor = JobExecutor.get_instance("slurm")
      4 job = Job(
      5     JobSpec(
      6         name="test",
   (...)
     11     )
     12 )
---> 14 job_executor.submit(job)
     15 native_id = job.native_id
     16 print(native_id)

File /global/common/software/matgen/rosen/miniconda/envs/test2/lib/python3.10/site-packages/psij/executors/batch/batch_scheduler_executor.py:215, in BatchSchedulerExecutor.submit(self, job)
    212     self._set_job_status(job, JobStatus(JobState.QUEUED,
    213                                         metadata={'native_id': job.native_id}))
    214 except subprocess.CalledProcessError as ex:
--> 215     raise SubmitException(ex.output) from None
    217 self._queue_poll_thread.register_job(job)

SubmitException: sbatch: error: Job request does not match any supported policy.
sbatch: error: Batch job submission failed: Unspecified error
```

For the record, this is on Perlmutter at NERSC.

hategan:
This adds a test that sets various string properties of the job spec tree to some random value and then checks the generated submit script for that value.

It also fixes #416. It is unclear what causes #416 to happen, since there are no code changes in psij that are related. The one possible explanation is that pystache became more strict on processing keys and will not make a global lookup if the current dictionary is missing a key.

As a result of this test, a few more attributes that were missing in various executors are now added. It is somewhat unclear what the benefits of some of these additions are (PBS uses special queue names for reservations rather than requiring some reservation id attribute on the job, like Slurm or LSF, so arguably one does not need to handle the reservation_id attribute for PBS, but then it's probably better to do so for portability reasons.)

### DIFF
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -7,7 +7,7 @@ from abc import abstractmethod
 from datetime import timedelta
 from pathlib import Path
 from threading import Thread, RLock
-from typing import Optional, List, Dict, Collection, cast, TextIO, Union
+from typing import Optional, List, Dict, Collection, cast, Union, IO
 
 from .escape_functions import bash_escape
 from psij.launchers.script_based_launcher import ScriptBasedLauncher
@@ -268,7 +268,7 @@ class BatchSchedulerExecutor(JobExecutor):
 
     @abstractmethod
     def generate_submit_script(self, job: Job, context: Dict[str, object],
-                               submit_file: TextIO) -> None:
+                               submit_file: IO[str]) -> None:
         """Called to generate a submit script for a job.
 
         Concrete implementations of batch scheduler executors must override this method in
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -38,6 +38,8 @@ def _attrs_to_mustache(job: Job) -> Dict[str, Union[object, List[Dict[str, objec
     for k, v in job.spec.attributes._custom_attributes.items():
         ks = k.split('.', maxsplit=1)
         if len(ks) == 2:
+            # always use lower case here
+            ks[0] = ks[0].lower()
             if ks[0] not in r:
                 r[ks[0]] = []
             cast(List[Dict[str, object]], r[ks[0]]).append({'key': ks[1], 'value': v})
--- a/src/psij/executors/batch/cobalt.py
+++ b/src/psij/executors/batch/cobalt.py
@@ -1,7 +1,7 @@
 """Defines a JobExecutor for the Cobalt resource manager."""
 from datetime import timedelta
 from pathlib import Path
-from typing import Optional, Collection, List, Dict, TextIO
+from typing import Optional, Collection, List, Dict, IO
 import re
 import os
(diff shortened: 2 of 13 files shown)


andre-merzky (approved):
This looks really thorough - nice!

PR #420, opened 2023-09-20
Move the walltime parsing routine in the serialiation module to

hategan:
JobAttributes and make it public.


andre-merzky (approved):
Makes sense, that looks useful to expose indeed.

PR #421, opened 2023-10-02
fix `AttributeError: module 'flux' has no attribute 'jobspec'`

adi611:
Update `flux.py` to fix the following error: `AttributeError: module 'flux' has no attribute 'jobspec'`


andre-merzky (approved):
Oh, thank you for catching that!

Issue #314, opened 2022-10-27
local executor should check env value type


andre-merzky:
The following code results in an exception:
```py
#!/usr/bin/env python3

import psij

job = psij.Job(psij.JobSpec(executable='/bin/true', environment={'foo':1}))
ex = psij.JobExecutor.get_instance('local')
ex.submit(job)
status = job.wait()
```

The traceback is:

```sh
Traceback (most recent call last):
  File "/mnt/home/merzky/radical/radical.pilot.2/ve3/lib/python3.8/site-packages/psij/executors/local.py", line 216, in submit
    p.process = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
  File "/usr/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.8/subprocess.py", line 1624, in _execute_child
    env_list.append(k + b'=' + os.fsencode(v))
  File "/mnt/home/merzky/radical/radical.pilot.2/ve3/lib/python3.8/os.py", line 804, in fsencode
    filename = fspath(filename)  # Does type-checking of `filename`.
TypeError: expected str, bytes or os.PathLike object, not int

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./t.py", line 7, in <module>
    ex.submit(job)
  File "/mnt/home/merzky/radical/radical.pilot.2/ve3/lib/python3.8/site-packages/psij/executors/local.py", line 224, in submit
    raise SubmitException('Failed to submit job', exception=ex)
psij.exceptions.SubmitException: Failed to submit job
```

To some extent that's clearly a user error: env dicts are documented as `str:str` typed, and the user (me) ignored that.  And type checking is for many an anti-pattern in Python.  But, oh boy, this took *way* to long for me to debug as the error message is not really helpful, and I would like to save others from the same time sink.  I would thus propose to either type check the env dict on spec construction, or to simply cast all values to `str`-type.  The latter may still be incorrect in some cases, but is much easier to debug on user level.

hategan:
> I would thus propose to either type check the env dict on spec construction, or to simply cast all values to `str`-type. The latter may still be incorrect in some cases, but is much easier to debug on user level.

[...]

PR #425, opened 2023-10-23
Fix env type

andre-merzky:
This addresses #314: job's environment dict can now have string and int types - int types are cast to string types.


hategan:
I need to undraft this for a minute to test something.

andre-merzky:
@hategan : can I have your advice on the [`mypy` error](https://github.com/ExaWorks/psij-python/actions/runs/7530712791/job/20497714352?pr=425)?  I seem to be somewhat blind on why the `environment.setter` declaration does not kick in :-(

hategan:
> @hategan : can I have your advice on the [`mypy` error](https://github.com/ExaWorks/psij-python/actions/runs/7530712791/job/20497714352?pr=425)? I seem to be somewhat blind on why the `environment.setter` declaration does not kick in :-(

Could it be https://mypy.readthedocs.io/en/stable/common_issues.html#variance?

I.e., you cannot assign a `Dict[str, str]` to a `Dict[str, str | int]`. But `Mapping` is covariant, so a `Mapping[str, str]` can be assigned to a `Mapping[str, str | int]`.

andre-merzky:
> > @hategan : can I have your advice on the [`mypy` error](https://github.com/ExaWorks/psij-python/actions/runs/7530712791/job/20497714352?pr=425)? I seem to be somewhat blind on why the `environment.setter` declaration does not kick in :-(
> 
> Could it be https://mypy.readthedocs.io/en/stable/common_issues.html#variance?
> 
> I.e., you cannot assign a `Dict[str, str]` to a `Dict[str, str | int]`. But `Mapping` is covariant, so a `Mapping[str, str]` can be assigned to a `Mapping[str, str | int]`.

Thanks @hategan , that looks like a good read!

andre-merzky:
Thanks for the pointer to Mapping, that does not seem to do the trick either though - I don't get mypy to like this without too much code to be worth the trouble really.  I see two options: silence the mypy warning (ugh), or drop the idea of accepting `int` values.  At this point I am fine with the second option also.  Let me know what you think :-)

hategan:
My apologies. I think we are hitting an entirely different issue than the plain invariance that I quoted.

What seems to be happening, in this particular case, is that mypy takes the type of the property from the getter. I say "in this particular case" because the same thing is not happening with all the `Path | str` getters/setters. It would appear that mypy should support this (see https://github.com/python/mypy/issues/3004), but it is unclear to my why it breaks for the dict and not paths.

hategan:
> but it is unclear to my why it breaks for the dict and not paths.

It also breaks for paths. It's just that we haven't tested it. For me, `spec.directory = '/tmp'` results in a mypy error.

hategan:
> see [python/mypy#3004](https://github.com/python/mypy/issues/3004)

That issue was opened in 2017. I had incorrectly assumed that it was fixed not long after. However, it is still open.

andre-merzky:
> I slept on it and I think we should allow int values.

If you are ok with that approach I'm happy to go that route also. I certainly agree on the pros and cons.  Thanks!

hategan:
> > I slept on it and I think we should allow int values.
> 
> If you are ok with that approach I'm happy to go that route also. I certainly agree on the pros and cons. Thanks!

I think my vote is to be OK with said approach.

tests/test_job_spec.py
```diff
@@ -12,16 +12,14 @@ def _test_spec(spec: JobSpec) -> None:
 
 
 def test_environment_types() -> None:
-    with pytest.raises(TypeError):
-        _test_spec(JobSpec(environment={'foo': 1}))  # type: ignore
 
     with pytest.raises(TypeError):
-        _test_spec(JobSpec(environment={1: 'foo'}))  # type: ignore
+        _test_spec(JobSpec(executable='true', environment={1: 'foo'}))  # type: ignore
```
hategan, line 17: I would caution against assuming that `/bin` is in `PATH` or that `PATH` is used when launching executables.

tests/plugins1/_batch_test/test/test.mustache
```diff
@@ -41,9 +41,7 @@ for NODE in $(seq 1 1 "$PSIJ_TEST_BATCH_EXEC_COUNT"); do
...
+{{#job.spec.inherit_environment}}{{/job.spec.inherit_environment}}{{^job.spec.inherit_environment}}env --ignore-environment \{{/job.spec.inherit_environment}}{{#env}}
+export {{name}}="{{value}}" {{/env}}
 
-{{#job.spec.inherit_environment}}env \{{/job.spec.inherit_environment}}{{^job.spec.inherit_environment}}env --ignore-environment \{{/job.spec.inherit_environment}}{{#env}}
-{{name}}="{{value}}" \
-{{/env}}{{#psij.launch_command}}{{.}} {{/psij.launch_command}}
-
-echo "$?" > "{{psij.script_dir}}/$PSIJ_BATCH_TEST_JOB_ID.ec"
\ No newline at end of file
+echo "$?" > "{{psij.script_dir}}/$PSIJ_BATCH_TEST_JOB_ID.ec"
```
hategan, line 47: I' m trying to understand this change. It was:

```bash
export --ignore-environment \
    A='a' \
    b='b' \
    command
```

That is, `env` is used to set the variables and run the command.

It looks like now it will expand to
```bash
export --ignore-environment \
    export A='a'
    export B='b'
    command
```

Here `env` is used to set the first variable, although I think this should fail since env will try to run the `export` executable, which is a bash builtin, so it will fail. Even if not, it will only set the first variable. Then the second and subsequent variables will be set, and the command run with the inherited environment (since the scope of the effects of  `env --ignore-environment` has ended) and all but the first variable set.

hategan (approved):
Looks good, but I'm not quite sure about the test executor template change.

tests/plugins1/_batch_test/test/test.mustache
andre-merzky: jeah, my 'fix' was not a fix apparently.  The problem is that I don't get the 

```sh
export --ignore-environment \
    A='a' \
    b='b' \
    command
```

but rather see 
```sh
export --ignore-environment \
    A='a' \
    
    b='b' \
    command
```

(See [mustache.txt](https://github.com/ExaWorks/psij-python/files/14311838/mustache.txt) for example).  I tried to get rid of the newline, but the escaping is a bit funny with mustache :-)  Reverted now, please let me know if you have any advise.

hategan: Indeed, and we weren't testing for multiple env vars.

I think the issue is that there is a newline after `{{#env}}`, which puts a newline before each env var line, and there is one after each env var, which adds another.

The following seems to work for me:
```
{{#job.spec.inherit_environment}}env \{{/job.spec.inherit_environment}}{{^job.spec.inherit_environment}}env --ignore-environment \{{/job.spec.inherit_environment}}{{#env}}
{{name}}="{{value}}" \{{/env}}
{{#psij.launch_command}}{{.}} {{/psij.launch_command}}
```

I.e., only add newline before each env var and before the launch command.

hategan: Perhaps it would be wise to fix in all executors.

andre-merzky: I don't think the other executors suffer from the same problem as they create individual `export` directives per env pair?

hategan (approved):
Thanks Andre.

PR #426, opened 2023-11-09
This updates the release script to ensure some sanity in the environm…

hategan:
…ent,

use a release branch to allow automatic updates of version numbers, and triggers a web-build (assuming workflow is properly configured)


andre-merzky (approved):
Looks clean and reasonable to me.  Thanks!

PR #428, opened 2023-11-16
Fix new key lookup

hategan:
Contrary to popular belief, python does not automatically expand `~` in pathlib.PosixPath objects.


andre-merzky (approved):
LGTM

PR #430, opened 2023-11-21
This sets the queue and project names if specified by `ExecutorTestPa…

hategan:
…rams`

since this is how the testing infrastructure is meant to pass these parameters.


andre-merzky (approved):
LGTM

PR #431, opened 2023-11-22
Some testing fixes

hategan:
This basically makes sure that `_get_executor_instance` is used when potentially dealing with queuing systems. This is done in order
to propagate various job attributes from testing.conf.

It also updates the web rendering library to recognize when a doc example has `_get_executor_instance(...)` and replace with something more palatable.


andre-merzky (approved):
Neat!

PR #433, opened 2024-01-07
Make cron test installation work even when the user's crontab is empty.

hategan:
It also fixes a broken test for the authentication key.


andre-merzky (approved):
Yeah, makes sense - thanks!

hategan:
> Yeah, makes sense - thanks!

Thank you and happy new year!

PR #436, opened 2024-01-12
Fix walltime float hours

### LINKED ISSUE #435: PBS walltime hour is being formatted as a float causing job submission failure
For example if I specify a `duration=timedelta(minutes=60)`, the resulting parameter in the submission script ends up being `#PBS -l walltime=1.0:0:0` which causes qsub to throw an error. I changed the `BatchSchedulerExecutor._format_duration` method to convert the hours to an int `int(d.total_seconds() // 3600)`.

hategan:
Fixes #435


andre-merzky (approved):
Yup!

PR #437, opened 2024-01-13
Add (lousy) detection of schedulers other than Slurm.

hategan:
This allows https://github.com/ExaWorks/psij-testing-service/pull/67 to show schedulers other than Slurm and enables actual testing of the scheduler on machines other than ones with Slurm.


andre-merzky (approved):
80:20 rule - looks good to me, we can improve once the 20% hit :-P

hategan:
Note: doc related error is being fixed by another PR.

PR #438, opened 2024-01-15
sphinxcontrib-applehelp 1.0.8 added a requirement for Sphinx 5.0, which

hategan:
breaks Sphinx 4.5 who doesn't pin the dep to something compatible with it.

It's even more silly because we don't generate applehelp output.


andre-merzky (approved):
If that's what we need `¯\_(ツ)_/¯`

andre-merzky (approved):
reapprove

PR #439, opened 2024-01-15
Added resource specs to PBS executor, which were somehow missing.

### LINKED ISSUE #434: Not able to get ResourceSpecV1 working with PBS
Just checking to see if I am just doing something wrong or not but I am not able to get a PBS select string any different than `1:ncpus=1` no matter what resource spec I give to the job. Is this a known issue?

hategan:
Also split PBS executor into pbs_classic (the one using `:ppn` to specify processes per node) and the modern pbspro/openpbs (using `select=:ncpus=`).

Fixes #434

### DIFF
--- a/src/psij-descriptors/pbs_descriptor.py
+++ b/src/psij-descriptors/pbs_descriptor.py
@@ -1,0 +1,11 @@
+from distutils.version import StrictVersion
+
+from psij.descriptor import Descriptor
+
+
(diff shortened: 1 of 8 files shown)


andre-merzky (approved):
I did not do a thorough review but also don't want to hold this up any longer.  From what I have seen this looks sensible and clean.  This will need to be tested on the respective schedulers - but I think a merge will only make tests more likely - so IMHO this is good to go.

PR #441, opened 2024-01-31
The process count, if not specified, would be empty. Most batch sched…

hategan:
…ulers

interpret this using a default of 1. However, the batch test scheduler uses the empty env var as a parameter to `seq`, which may lead to failures.


andre-merzky (approved):
LGTM

PR #442, opened 2024-02-02
Use `assert_completed` when possible and also change `assert_completed`

hategan:
to not complain of a null spec when checking an attached job.


andre-merzky (approved):
LGTM

PR #443, opened 2024-02-02
Added a way to specify test filters for custom attributes.

hategan:
The motivation here is that, on Perlmutter, it's a wise idea to run most tests on shared nodes (since we don't want to burn too fast through a small allocation). However, some tests, such as tests that verify that multi-node jobs run on multiple nodes, will fail since the shared node mode (--qos=shared) does not allow for multi-node jobs. So there needs to be a way to use a different --qos parameter for certain tests.


andre-merzky (approved):
neat!

PR #444, opened 2024-02-07
Detect launcher failure in the batch scheduler executors. Somehow this

hategan:
was only done in the local executor.


andre-merzky (approved):
makes sense of course...

PR #445, opened 2024-02-07
Fixed handling of env vars in Slurm. The existing template made some

hategan:
assumptions about how the process worked that were incorrect. Specifically, you cannot issue multiple `#SBATCH --export` directives and expect them to add up. Only one should be used.


andre-merzky (approved):
LGTM

PR #446, opened 2024-02-08
Fix slurm nodefiles

hategan:
`scontrol show hostnames` only shows one line per node, no matter how many processes we requested. In order to bring this in line with the uniform PBS way of doing things (one line per process in the nodefile, possibly with duplicate entries), we need to do some manual work.

This also further fixes environment variable processing. Variables were set before the `#SLURM --export` directives, which means that the latter were effectively ignored by Slurm which stop processing once anything non-comment is found in the script.

### DIFF
--- a/src/psij/executors/batch/slurm/slurm.mustache
+++ b/src/psij/executors/batch/slurm/slurm.mustache
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+
 {{#job.name}}
 #SBATCH --job-name="{{.}}"
 {{/job.name}}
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
Duh!  on the `export` fix, sorry for missing that!
Thanks on the nodefile mangling.

hategan:
> Duh! on the `export` fix, sorry for missing that! Thanks on the nodefile mangling.

For reference, I checked the templates for the other executors and they don't seem to have a similar issue, but that's likely because they deal with env vars somewhat differently.

PR #448, opened 2024-02-10
Files used by jobs need to be on a shared FS. This is typically the case

hategan:
except that the ci_runner, when testing various branches, does the cloning in `/tmp`. One alternative solution would be to ensure this is moved to the home dir.

This was basically why `test_prelaunch` and `test_nodefile` failed.

### DIFF
--- a/tests/_test_tools.py
+++ b/tests/_test_tools.py
@@ -67,6 +67,7 @@ def _deploy(path: Union[Path, str]) -> Iterator[Path]:
         path = Path(path)
     with tempfile.NamedTemporaryFile(dir=Path.home() / '.psij' / 'test', delete=False) as df:
         try:
+            df.close()
             shutil.copyfile(path, df.name)
             yield Path(df.name)
         finally:
--- a/tests/_test_tools.py
+++ b/tests/_test_tools.py
@@ -1,6 +1,10 @@
+import os
+import shutil
+import tempfile
+from contextlib import contextmanager
 from datetime import timedelta
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union, Iterator
 
 from executor_test_params import ExecutorTestParams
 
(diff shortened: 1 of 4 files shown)


tests/_test_tools.py
```diff
@@ -53,3 +57,17 @@ def _get_executor_instance(ep: ExecutorTestParams, job: Optional[Job] = None) ->
...
     return JobExecutor.get_instance(ep.executor, url=ep.url)
+
+
+@contextmanager
+def _deploy(path: Union[Path, str]) -> Iterator[Path]:
+    # Copies `path` to a directory assumed to be on a shared FS (~/.psij/test) and
+    # returns the resulting path
+    if isinstance(path, str):
+        path = Path(path)
+    with tempfile.NamedTemporaryFile(dir=Path.home() / '.psij' / 'test', delete=False) as df:
```
andre-merzky, line 68: Why the `delete=False` and `finally` instead of `delete=True` (which is the default anyway):
```py
with tempfile.NamedTemporaryFile(dir=Path.home() / '.psij' / 'test') as df:
    shutil.copyfile(path, df.name)
    yield Path(df.name)
```

andre-merzky (approved):
left a comment, otherwise looks good to me - thanks!

tests/_test_tools.py
```diff
@@ -67,6 +67,7 @@ def _deploy(path: Union[Path, str]) -> Iterator[Path]:
         path = Path(path)
     with tempfile.NamedTemporaryFile(dir=Path.home() / '.psij' / 'test', delete=False) as df:
         try:
+            df.close()
```
andre-merzky, line 70: Ah! :-D

tests/_test_tools.py
hategan: Good point and thanks for commenting on this, since I forgot to close the file. With `delete=True` it deletes on close. So if the scenario is to manipulate the file outside of python, you'd close it and pass the name around instead. Without `delete=False`, the file will be deleted as soon as it is closed. The need to close is because on some platforms (Windows) the file cannot be opened concurrently. The question that the Python documentation doesn't answer is whether, if the file is deleted, the name property would still be valid. The wording of the documentation suggests that this may not be the case.

andre-merzky: Yeah, makes sense now, thanks!

PR #451, opened 2024-02-24
Checking existence and then deleting is not atomic, so try to delete and

hategan:
ignore FileNotFoundError instead

See test_executor::test_attach2[slurm:srun on [https://testing.psij.io/run.html?...](https://testing.psij.io/run.html?site_id=perlmutter.nersc.gov&run_id=2024022406-3ac2618e18df98e9eedf6f962587213e5f3e24a41076615c4973d544da9972ea#branch-main) for the failure.


andre-merzky (approved):
LGTM - thanks!

PR #452, opened 2024-03-05
Use installed packages in ci

hategan:


andre-merzky (approved):
Nice catch :-)

hategan:
> Nice catch :-)

Wouldn't need catching if I did it right the first time!

PR #453, opened 2024-03-05
The entry is there, but is set to `false` when slurm is not detected.

hategan:


tests/conftest.py
```diff
@@ -420,7 +420,7 @@ def _discover_environment(config):
             logger.warning('Cannot get git repository information.')
     try:
         env['has_slurm'] = shutil.which('sbatch') is not None
-        if 'has_slurm' not in env:
+        if 'has_slurm' not in env or not env['has_slurm']:
```
andre-merzky, line 423: Wouldn't `if not env.get('has_slurm')` be simpler?  But either way, fix looks correct to me!

andre-merzky (approved):
LGTM!

tests/conftest.py
hategan: Yes, yes it would, of course, exactly because we know that the entry is there

PR #454, opened 2024-03-05
Documented custom attributes a bit better

### LINKED ISSUE #415: Proposal: Allow JSON-serializable format for `duration` in `JobAttributes`
Currently, `JobAttributes` takes a `duration` kwarg that must be of type `datetime.timedelta`. While this makes sense, it is posing some challenges for me because `datetime.timedelta` is not JSON serializable, which is problematic for how some workflow engines operate. It would be much easier to handle if `duration` could be supplied as an alternate type. Perhaps the easiest route would be a `str` formatted as `DD:HH:SS` or even easier, an `int` or `float` in a pre-defined unit (e.g. minutes) that would be transformed to a date time object when the `JobAttributes` class isinstantiated.

What I'm thinking is something like:

```python
self.duration = duration if isinstance(duration, timedelta) else timedelta(minutes=duration)
```

hategan:
Fixes #415

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ psi_j_python.egg-info/
 venv*
 .venv*
 build/
+docs/.web-build
+web-build/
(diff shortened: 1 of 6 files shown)


andre-merzky (approved):
Great!

PR #458, opened 2024-03-11
Use packaging.versions insted of distutil.version, since the latter

hategan:
is deprecated in Python 3.1x ish.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ venv*
 build/
 docs/.web-build
 web-build/
+.packages/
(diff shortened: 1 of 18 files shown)


andre-merzky (approved):
Thanks for doing this tedious thing!

`StrictVersion` now only mentioned in `docs/development/tutorial_add_executor.rst` and probably should be removed there, too?

hategan:
> Thanks for doing this tedious thing!

Wasn't too bad.

> 
> `StrictVersion` now only mentioned in `docs/development/tutorial_add_executor.rst` (and `docs/conf.py`) and probably should be removed there, too?

Thank you for grepping for that and apologies that I didn't do it in the first place. Should be fixed now.

PR #459, opened 2024-03-20
Added `account` job attribute to use for accounting/billing.

### LINKED ISSUE #456: Account vs. project
PBS distinguishes between an account (`-A` option) and a project (`-P` option). At least one system (improv@lcrc.anl) needs the account to be set to what the allocation system calls a "project". The current executor, on the other hand, maps `spec.project` to `-P`.

Anyway, we have two things, and only one of them is supported by the job spec, so we probably need to decide whether to amend the specification to account for both.

@andre-merzky, thoughts?

hategan:
Deprecate the `project` attribute and re-direct it to `account`.

Fixes #456.

### DIFF
--- a/QuickStart.md
+++ b/QuickStart.md
@@ -58,7 +58,7 @@ def make_job():
     spec.arguments = ['HELLO WORLD!']
     
     # set project name if no default is specified
-    # spec.attributes.project_name = <PROJECT_NAME>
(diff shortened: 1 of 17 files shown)


andre-merzky (approved):
Dammit, I missed the review on this one, sorry!  Looks good to me.

hategan:
No worries. Thanks for getting to it.

PR #460, opened 2024-04-02
When testing, always put the src dir before .packages in PYTHONPATH. 

hategan:
We don't want a dependency to install something that may have the same name as something in psij (including possibly a different version of psij itself) and hijack things.

@andre-merzky: We have a bit of a circular dependency: psij -> rp -> psij. This should be fine, since psij -> some_pilot_system -> psij is one of the intended scenarios. We may want to discuss whether such dependencies are necessarily package dependencies.

### DIFF
--- a/tests/ci_runner.py
+++ b/tests/ci_runner.py
@@ -164,8 +164,8 @@ def run_branch_tests(conf: Dict[str, str], dir: Path, run_id: str, clone: bool =
 
     cwd = (dir / 'code') if clone else Path('.')
     env = dict(os.environ)
-    env['PYTHONPATH'] = str(cwd.resolve() / '.packages') \
-        + ':' + str(cwd.resolve() / 'src') \
+    env['PYTHONPATH'] = str(cwd.resolve() / 'src') \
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):

andre-merzky:
> We don't want a dependency to install something that may have the same name as something in psij (including possibly a different version of psij itself) and hijack things.
> 
> @andre-merzky: We have a bit of a circular dependency: psij -> rp -> psij. This should be fine, since psij -> some_pilot_system -> psij is one of the intended scenarios. We may want to discuss whether such dependencies are necessarily package dependencies.

Oh, indeed - I did not even think of that...  Yes, ultimately it's an intended use case.  But RP is not a hard dependency for psij-python, so are you referring to dependencies in a CI context?

hategan:
> Oh, indeed - I did not even think of that... Yes, ultimately it's an intended use case. But RP is not a hard dependency for psij-python, so are you referring to dependencies in a CI context?

Yes. But this is likely to resurface in the form of warnings about multiple executors being found if PSIJ is installed with RP.

PR #464, opened 2024-04-08
typeguard update

### LINKED ISSUE #461: Dependency on typeguard 2.12 conflicts with other packages.
At present psij-python has typeguard 2.12 as a requirement (see requirements.txt). This causes conflicts with other packages that have moved on to typeguard > 3.0.1. However the required changes seem minimal. 

Using typeguard 2.12 you would write:
```python
from typeguard import check_argument_types
...
def function(...):
    ...
    assert check_argument_types()
    ...
```
Using typeguard > 3.0.1 this becomes:
```python
from typeguard import typechecked
...
@typechecked
def function(...):
```
So far I have found 3 locations where this change needs to be made:

- src/psij/job_attributes.py
- src/psij/job_spec.py
- src/psij/resource_spec.py

Given that the require modifications seem minimal, and that the old typeguard 2.12 is incompatible with the post typeguard 3.0.1 versions, would it make sense to make these changes and update the version of typeguard that psij-python depends on?

andre-merzky:
This fixes #461

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 filelock~=3.4
 psutil~=5.9
 pystache>=0.6.0
-typeguard~=2.12
+typeguard>=3.0.1
 typing-compat
 packaging~=24.0
--- a/src/psij/job_attributes.py
+++ b/src/psij/job_attributes.py
@@ -43,7 +43,6 @@ class JobAttributes(object):
 
         All constructor parameters are accessible as properties.
         """
-
         self.duration = duration
         self.queue_name = queue_name
         self.project_name = project_name
--- a/src/psij/job_attributes.py
+++ b/src/psij/job_attributes.py
@@ -2,7 +2,7 @@ import re
 from datetime import timedelta
 from typing import Optional, Dict
 
-from typeguard import check_argument_types
+from typeguard import typechecked
 
 
 _WALLTIME_FMT_ERROR = 'Unknown walltime format: %s. Accepted formats are hh:mm:ss, ' \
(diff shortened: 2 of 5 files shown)


hategan:
Annoying. @typechecked is broken.

But. It seems that if you add @typechecked to the class rather than `__init__`, mypy will not complain. I suspect it derives the type of `Cons()` using `__init__` and some heuristics.

andre-merzky:
> Annoying. `@typechecked` is broken.
> 
> But. It seems that if you add `@typechecked` to the class rather than `__init__`, mypy will not complain. I suspect it derives the type of `Cons()` using `__init__` and some heuristics.

Yeah, it's broken.  Given that I guess that the class decorating approach is the way of least pain indeed...  Thanks!

hategan:
For reference, `make checks` does not report any errors on my machine and tests on testing.psij.io seem to pass on this branch.

andre-merzky:
> For reference, `make checks` does not report any errors on my machine and tests on testing.psij.io seem to pass on this branch.

Yeah, same here - tests pass fine locally, docs build all right, but github CI still complains...

hategan:
> > For reference, `make checks` does not report any errors on my machine and tests on testing.psij.io seem to pass on this branch.
> 
> Yeah, same here - tests pass fine locally, docs build all right, but github CI still complains...

I think we can merge. If we run into issues building the docs after, we can address it then.

hategan:
By the way, I approve of this PR :)

mtitov (approved):
LGTM!

hategan:
The failing tests are due to RP installing an older version of PSI/J as a dependency, which conflicts with the version being tested. While we should address the "circular" dependency (there is little sense to use PSI/J to submit to PSI/J via RP), this needs not happen now since RP is an optional install.

Issue #468, opened 2024-05-06
Deprecation warning for "project_name" is always triggered


Andrew-S-Rosen:
With psij-python 0.9.7:

```python
from psij.job_attributes import JobAttributes
JobAttributes()
```

Returns:

```
The "project_name" attribute is deprecated. Please use the "account" attribute instead.
```

The same thing happens when `account` is directly provided. Presumably, this is because "project_name" is being set in the init? I haven't looked into this deeply.

hategan:
Fixed by #469.

PR #469, opened 2024-05-06
Only trigger warning about `project_name` if it's being set to something

hategan:
that isn't `None`.

Thanks @Andrew-S-Rosen for reporting this (#468).


andre-merzky (approved):
Yup.

hategan:
Thank you!

PR #471, opened 2024-05-13
Move filelock requirement from requirements.txt to requirements-tests…

### LINKED ISSUE #465: Dependency `filelock` move to requirements-tests
Can we move `filelock` from [requirements.txt](https://github.com/ExaWorks/psij-python/blob/main/requirements.txt) to [requirements-tests.txt](https://github.com/ExaWorks/psij-python/blob/main/requirements-tests.txt)

hategan:
….txt

since it appears to be the only place where it is actually used.

Fixes #465

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -7,3 +7,4 @@ pytest >= 6.2.0
 requests >= 2.25.1
 pytest-cov
 pytest-timeout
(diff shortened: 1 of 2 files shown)


mtitov (approved):
LGTM! Thanks!

PR #486, opened 2024-07-01
`squeue` fails if a job id is passed to `-j` for a job that is not in…

hategan:
… the

queue (any more). Because we cannot guarantee that the query intervals are smaller than the intervals at which jobs get purged from the queue, failed `squeue` invocations can happen.

This change replaces the `-j` option with `--me`, which lists all jobs for a user. This has the potential of listing more jobs than necessary (since a user can have multiple PSI/J instances managing jobs).


andre-merzky (approved):
I don't think the overhead of listing additional jobs will matter much in like 99.999% of the cases.  Looks good to me :-)

hategan:
Thank you @andre-merzky!

PR #487, opened 2024-07-01
Use uid for batch-test scheduler files stored in `/tmp` in order to a…

hategan:
…void

issues when multiple users use it.


andre-merzky (approved):
Yup, that makes sense.  Thanks!

PR #495, opened 2025-01-09
On Frontier, the node count is mandatory. We should probably discuss

hategan:
having one node as a default and specify that in the spec. In the mean time, add one node as default for slurm.


hategan:
On a second thought, I will scrap this and work on a more general solution.

PR #502, opened 2025-02-10
Fix thread "leak"

### LINKED ISSUE #501: QueuePollThread instance leak
A _QueuePollThread instance is created for each instance of BatchSchedulerExecutor (or any of its subclasses). The typical usage scenario is that a small number of executor instances would be reused for multiple jobs. However, one can envision potential cases when many executor instances are needed (or at least used), such as when the tests are run. This is wasteful, since threads are kept for the entire lifetime of the python VM even though they are only used momentarily and the executors that use them are not referenced any more. 

In particular, when running the tests under screen on expanse.sdsc.edu, by the time the last few tests get to run, thread count limits are reached, causing the tests to fail.

hategan:
This fixes #501 by replacing the executor property of _QueuePollThread by weakref and shutting down the thread when
the executor is collected. The weak ref is necessary to avoid a reference loop which would prevent a `__del__` based solution from working.

### DIFF
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -3,6 +3,7 @@ import os
 import subprocess
 import time
 import traceback
+import weakref
 from abc import abstractmethod
 from datetime import timedelta
 from pathlib import Path
@@ -639,7 +640,8 @@ class _QueuePollThread(Thread):
         self.name = name
         self.daemon = True
         self.config = config
-        self.executor = executor
+        self.done = False
+        self.executor = weakref.ref(executor, self._stop)
         # native_id -> job
         self._jobs: Dict[str, List[Job]] = {}
         # counts consecutive errors while invoking qstat or equivalent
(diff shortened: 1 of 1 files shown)


src/psij/executors/batch/batch_scheduler_executor.py
```diff
@@ -639,45 +640,51 @@ def __init__(self, name: str, config: BatchSchedulerExecutorConfig,
...
+        if exec is None:
+            return
         with self._jobs_lock:
             if len(self._jobs) == 0:
                 return
             jobs_copy = dict(self._jobs)
         logger.info('Polling for %s jobs', len(jobs_copy))
         try:
-            out = self.executor._run_command(self.executor.get_status_command(jobs_copy.keys()))
+            if exec:
```
andre-merzky, line 671: `if exec` is not needed, that is checked above.

andre-merzky (commented):
Sorry for late response - LGTM!

src/psij/executors/batch/batch_scheduler_executor.py
hategan: Ah, thank you. Will issue a small PR for that.

hategan:
> Sorry for late response - LGTM!

No worries. I thought you were away. Glad you took a look.

PR #504, opened 2025-02-14
Adds a `multi_node_queue_name` ci config option

hategan:
If specified, this option allows multi-node tests to run using a different queue than the one in `queue_name`. The wisdom of this is that on certain systems it makes sense to use more restrictive but less expensive queues by default, and only use more expensive, multi-node queues when strictly needed. In a sense, this could be hacked using `custom_attributes` and test filters, which some deployments do, but it is awkward and difficult to set up in a way that predicts all future multi-node tests.

When merged, the new `ci_runner.py` will automatically be used by all deployed tests. It was modified to not pass config
options that are not set, and all such options should have defaults in the `conftest.py` arg parser, both old and new.

This addition will affect tests that are currently deployed as follows (new and old refer to post and pre merging this PR):
1. old_clone -> new_pr: OK (new PRs work with the old options)
2. old_clone -> old_pr: OK (has old config and new ci_runner does not pass missing options)
3. new_clone -> old_pr: Not OK (will have to update PR - only one ATM; will pass new option which is not recognized)
4. new_clone -> new_pr: OK by construction

### DIFF
--- a/testing.conf
+++ b/testing.conf
@@ -116,6 +116,14 @@ queue_name =
 
 
 #
+# If a different queue is used for multi-node jobs, specify it here. If empty,
+# queue_name is used for both single-node and multi-node jobs.
+#
+
+multi_node_queue_name =
+
+
+#
 # If you need a project/account for billing purposes, enter it here.
 #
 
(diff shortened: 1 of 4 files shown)


hategan:
> This looks reasonable to me - thanks!

Thank you!

PR #506, opened 2025-02-19
This was meant to be part of PR #504

hategan:


hategan:
Will fast-track this given that it is an obvious omission.

PR #507, opened 2025-02-19
Add "quiet" flag for mpirun to avoid messages interfering with output capture.

hategan:
mpirun, on some systems, can print "helpful messages". Except they aren't
helpful for us trying to run automated systems with predictable behavior.


andre-merzky (approved):
LGTM

PR #508, opened 2025-02-21
Wait for job to complete in `test_list`

hategan:
On some machines, the tests use a fast queue for simple jobs that only allow one (or a small number of) job(s) to be queued at a time. If we don't wait for the job to complete, the next job may lead to a violation of queue policies.


andre-merzky (approved):
Annoying (because slow), but I don't see a better way.  Thanks!

hategan:
We could, in principle, cancel it, but the incremental gain is small.

PR #509, opened 2025-02-24
Better extraction of stdout/stderr streams from mpirun

hategan:
On delta.ncsa, mpirun produces various diagnostic output along with the application stdout. This output is not suppressed by the `-v` parameter to mpirun. Instead, we can use output tagging, which marks each application output line with `[jobid, rank]<stdxxx>:...`.

The PR also fixes launcher error reporting as well as adds a test for stderr redirect.


andre-merzky (approved):
LGTM!

hategan:
Thank you @andre-merzky as usual. And sorry for all the volunteer work :)

PR #512, opened 2025-03-11
Allow passing a vanilla JobExecutorConfig to batch scheduler executors

### LINKED ISSUE #511: Setting work_directory in abstract class JobExecutorConfig breaks SlurmExecutor with no attribute 'initial_queue_polling_delay'
Running with this executor doesn't work:
```
    JobExecutor.get_instance(
        "slurm",
        config=JobExecutorConfig(
            launcher_log_file=Path(launcher_log_file),
            work_directory=Path(work_directory),
        ),
    )
```

failing with this error:
```
AttributeError: 'JobExecutorConfig' object has no attribute 'initial_queue_polling_delay'
```

I can fix it by using a `BatchSchedulerExecutorConfig`, but I'm pretty sure this isn't the intended use of the API, because of the obscure import path

```
from psij.executors.batch.batch_scheduler_executor import BatchSchedulerExecutorConfig

    JobExecutor.get_instance(
        "slurm",
        config=BatchSchedulerExecutorConfig(
            launcher_log_file=Path(launcher_log_file),
            work_directory=Path(work_directory),
        ),
    )
```

For now I am indeed using `BatchSchedulerExecutorConfig`, but I feel like I'm going off-piste here.

hategan:
When passing a plain `JobExecutorConfig` to a batch scheduler executor, the executor attempts to access various batch scheduler specific properties, which do not exist (see #511).

This PR checks if a vanilla `JobExecutorConfig` is passed to a batch scheduler executor and replaces it with a `BatchSchedulerExecutorConfig` while copying the relevant properties.

Fixes #511.

### DIFF
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -113,6 +113,13 @@ class BatchSchedulerExecutorConfig(JobExecutorConfig):
         if 'PSIJ_BATCH_KEEP_FILES' in os.environ:
             self.keep_files = True
 
+    @classmethod
+    def _from_config(cls, config: JobExecutorConfig) -> 'BatchSchedulerExecutorConfig':
+        new = cls()
+        new.work_directory = config.work_directory
+        new.launcher_log_file = config.launcher_log_file
+        return new
+
 
 class InvalidJobStateError(Exception):
     """An exception that signals that a job cannot be cancelled due to it being already done."""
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
This looks correct and complete to me - thanks for the fast fix!

One thing to keep in mind is that `_from_config` needs updating whenever the `JobConfig` class changes.  Probably not worth to get clever at this point, we can revisit once we stumble over this again (if ever).

hategan:
> One thing to keep in mind is that `_from_config` needs updating whenever the `JobConfig` class changes. Probably not worth to get clever at this point, we can revisit once we stumble over this again (if ever).

Indeed. I went with a simple and static solution. We could do it by iterating over properties, but I'm not sure the gains are worth the loss in clarity.

I'm going to wax philosophical, but does any static language have a mechanism to deal with something like this?

tesujimath:
Great, thank you!  And thanks for the quick fix.  Looking forward to the next release at which time I will switch back to vanilla `JobExecutorConfig`. 😁

PR #515, opened 2025-04-11
Fixed incorrect order of parameters in BatchSchedulerExecutorConfig

### LINKED ISSUE #514: BatchSchedulerExecutorConfig constructor propagates args incorrectly
`BatchSchedulerExecutorConfig` instantiates the superclass with args, but its order seems wrong.

https://github.com/ExaWorks/psij-python/blob/d49a2a406ea6be0a3036a15d13c1bd2fc7e459b8/src/psij/executors/batch/batch_scheduler_executor.py#L108

https://github.com/ExaWorks/psij-python/blob/d49a2a406ea6be0a3036a15d13c1bd2fc7e459b8/src/psij/job_executor_config.py#L14-L15

When both are specified, job scripts are created in unexpected folder and the launcher log may be dumped into a folder, which crashes the workflow.

hategan:
Fixes #514

### DIFF
--- a/src/psij/executors/batch/batch_scheduler_executor.py
+++ b/src/psij/executors/batch/batch_scheduler_executor.py
@@ -105,7 +105,7 @@ class BatchSchedulerExecutorConfig(JobExecutorConfig):
             Whether to keep submit files and auxiliary job files (exit code and output files) after
             a job has completed.
         """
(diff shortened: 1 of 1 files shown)


andre-merzky (approved):
Thanks for the quick fix!

PR #517, opened 2025-04-30
Add `memory` property to `ResourceSpecV1`

hategan:
Some deployments require a memory specification when using nodes in non-exclusive mode. This PR adds that feature.

### DIFF
--- a/src/psij/executors/batch/lsf/lsf.mustache
+++ b/src/psij/executors/batch/lsf/lsf.mustache
@@ -32,6 +32,10 @@
 #BSUB -gpu num={{.}}/task
     {{/gpu_cores_per_process}}
 
+    {{#memory}}
+#BSUB -M {{memory_kb}}KB
+    {{/memory}}
+
 {{/job.spec.resources}}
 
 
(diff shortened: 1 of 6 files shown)


andre-merzky (approved):
Thanks Mikhail - that all looks good.  The only comment (and I'm not sure it warrants any change) is that `kb` is a bit arbitrary and small - would any application ever specify memory in that granularity?  But I also agree it's better to err on the site of finer granularity than on the coarser end...

Approving!

andre-merzky:
> > Thanks Mikhail - that all looks good. The only comment (and I'm not sure it warrants any change) is that `kb` is a bit arbitrary and small - would any application ever specify memory in that granularity? But I also agree it's better to err on the site of finer granularity than on the coarser end...
> > Approving!
> 
> The resource spec has it in bytes. But some of the schedulers don't use granularity lower than KB (i.e., `--mem 1` means 1 KB and `--mem 1b` is not allowed). The templates explicitly have the unit for clarity.

Thanks for clarifying - as the smallest common denominator, the unit does make sense.

PR #522, opened 2025-05-28
Add executor for NEC’s job scheduler (NQSV)

tat-ohmura:
This pull request adds an executor that supports NQSV (https://www.nec.com/en/global/solutions/hpc/articles/tech08.html). This is a proprietary job scheduler developed by NEC and is used in multiple large-scale HPC systems in Japan.
To wait for the completion of a job, PSI/J currently polls the status of the job. Since NQSV offers a dedicated command (qwait ) to wait for a job, we have implemented job completion detection using qwait instead of polling.
We hope to discuss and refine the implementation based on your feedback. Once the PR is reviewed and approved, we will update the documentation accordingly.

### DIFF
--- a/src/psij-descriptors/nqsv_descriptor.py
+++ b/src/psij-descriptors/nqsv_descriptor.py
@@ -1,0 +1,5 @@
+from packaging.version import Version
+from psij.descriptor import Descriptor
+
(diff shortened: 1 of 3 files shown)


hategan (approved):
This looks good. 

Codespell flagged a misspelled word in a comment, so I won't merge now if you want to fix that, but it's not important.

PR #524, opened 2025-07-02
Typing fixes

hategan:
The newest version of mypy found some issues, one of which was an actual bug.


hategan:
> Thanks!

Thank you!

PR #527, opened 2025-07-10
Updated Typeguard requirement to >=4.0.0. While Typeguard 3.x.x is 2 …

hategan:
…years

old or so, it is unfortunately preventing PSI/J from running (see https://github.com/agronholm/typeguard/issues/318 and https://github.com/ExaWorks/psij-python/issues/526).


mtitov (approved):
LGTM! Thank you!

